PR #57, opened 2020-09-23
Zen 355 styles fixes

lancetipton:
**Ticket [ZEN-355](https://jira.simpleviewtools.com/browse/ZEN-355)** 

> **Notes**
> This PR is directly tied to [this PR](https://github.com/simpleviewinc/keg-hub/pull/38)
>   * They should be merged at the same time
>
> Originally this ticket was to add a `react-helmet`
>   * I could not get that library to not de-dupe styles that were already added
>   * I knew it had to be possible, so I wrote it myself
>
> Because of this, the jira ticket and this pull request don't really line up
>   * This should be the last styling change ticket. At least I hope it is
>   * I'm tired of dealing with it

## Context
* Events-Force is defining their styles in CSS
* Their clients can define custom CSS as well
* Currently, the tap, using re-theme, is only compatible with JSON style objects
* We need to be able to pull styles from stylesheets already applied to the DOM and apply them to keg-components

## Goal
* Use the already added className prop, to allow applying styles useing the dom
  * This should only happend when in a web context
  * Native context, the style attribute should still be used
  * Ensure no changes need to be added to taps when defining styles
  * Ensure styles are not duplicated on the Dom
    * This is gold right here. I can't find any other repo that does this
    * It puts ReTheme on another level in my opinion 
* Taps can now do the following
  * Use `.css` styles applied to the dom and classNames **(web only)**
  * Use the style attribute, and still get all the benefits of the Dom
    * This works cross-platform

## Updates
* `package.json`
  * Fixed issue with start command
* `src/assets/icons/evf/elements/dayToggleLeft.js` && `src/assets/icons/evf/elements/dayToggleRight.js`
  * Clean up a little bit
  * Added memo hook
* `src/components/dates/dayToggle.js`
  * Fixed some styling issues
    * Updated to use `styles` instead of `style`
* `src/components/dates/updateDayButton.js`
  * Cleaned up styles
* `src/components/sessions.js`
  * Added ScrollView from `Keg-Components`
* `src/containers/gridContainer.js`
  * Added `?` on style paths
* `src/containers/rootContainer.js`
  * Added direct import of `sessionsContainer`
    * Was causing issues with mobile
* `src/theme/components/button/*`
  * **Most of these changes are to fix styling on mobile**
  * Cleaned up a number issues with style sheet parsing
  * Updated to work with changes made to `ReTheme`
* `./tap.js`
  * Building on mobile throws errors when using `const path = require('path')`
  * Updated to use relative paths based on the packager

## Testing

> **Notes**
> This is the same tests as [this PR](https://github.com/simpleviewinc/keg-hub/pull/38)
> So if you already did the tests for that one, they you have test this one

* Run Docker package => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:keg-mobile-build`
  * Ensure it runs
  * Open the inspector and check out the `Dom Tree`
    * Styles should no longer be applied to the elements with the style attribute
      * [All View / Text components should be using a class](http://snpy.in/MwwBpH)
      * [Some will still have styles ( Animated elements / Some primitives )](http://snpy.in/TuQOGY)
      * [Some will have buttons and styles for animations (buttons)](http://snpy.in/3d2mAV)
* Test Docker package => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:keg-mobile-build test`
  * Ensure all tests pass

### DIFF
--- a/src/theme/components/button/evfButton.js
+++ b/src/theme/components/button/evfButton.js
@@ -28,12 +28,12 @@ const defaultButtonStyles = {
   },
   content: {
     $xsmall: {
-      letterSpacing: '0.105em',
+      letterSpacing: 0.105,
       fontWeight: 'bold',
       color: colors.white,
       fontSize: 13,
       paddingHorizontal: 8,
-      lineHeight: '18px',
+      lineHeight: 18,
     },
     $small: {
       fontSize: 15,
--- a/src/theme/components/button/evfButton.js
+++ b/src/theme/components/button/evfButton.js
@@ -2,15 +2,21 @@ import { colors } from '../../colors'
 
 const topLeftCornerStyle = {
   main: {
-    width: 25,
-    height: 25,
-    backgroundColor: colors.white,
-    position: 'absolute',
-    overflow: 'hidden',
-    left: '-13px',
-    top: '-13px',
-    zIndex: 9999,
-    transform: [{ rotate: '45deg' }],
+    $web: {
+      left: '-13px',
+      top: '-13px',
+    },
+    $all: {
+      left: -13,
+      top: -13,
+      width: 25,
+      height: 25,
+      backgroundColor: colors.white,
+      position: 'absolute',
+      overflow: 'hidden',
+      zIndex: 9999,
+      transform: [{ rotate: '45deg' }],
+    },
   },
 }
 
--- a/tap.js
+++ b/tap.js
@@ -2,8 +2,8 @@
 
 
 const pathToEntryPoint = process.env.TEST_BUILD 
-  ? 'apps/BuildTest.js'
-  : 'apps/Main.js'
+  ? '../../apps/BuildTest.js'
+  : '../../apps/Main.js'
 
 module.exports = {
   name: 'events-force-x5',
--- a/tap.js
+++ b/tap.js
@@ -1,12 +1,16 @@
 
-
-const findPath = process.env.PLATFORM === 'web'
+/**
+ * It would be better to use the node core module `path`
+ * But this file is imported into the the app rootContainer, which is frontend
+ * So instead if finds the path to the tap root based on the Platform
+ */
+const tapRootPath = process.env.PLATFORM === 'web'
   ? '.'
   : '../..'
 
 const pathToEntryPoint = process.env.TEST_BUILD 
-  ? `${findPath}/apps/BuildTest.js`
-  : `${findPath}/apps/Main.js`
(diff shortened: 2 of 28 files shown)


tap.js
```diff
@@ -1,8 +1,12 @@
-const path = require('path')
+
+
+const findPath = process.env.PLATFORM === 'web'
+  ? '.'
+  : '../..'
```
mpcarolin, line 5: can you add a comment explaining why this is necessary? Sounds like it has something to do with Metro?

mpcarolin (approved):
Passed testing and looks good to me. Added one non-blocking suggestion.

Still reviewing the other PR.

src/theme/components/button/evfButton.js
```diff
@@ -2,15 +2,21 @@ import { colors } from '../../colors'
...
-    width: 25,
-    height: 25,
-    backgroundColor: colors.white,
-    position: 'absolute',
-    overflow: 'hidden',
-    left: '-13px',
-    top: '-13px',
-    zIndex: 9999,
-    transform: [{ rotate: '45deg' }],
+    $web: {
```
daniel112, line 5: marking this file for reference.
The  `Default` EvfButton is not getting the right background color (it's getting all the other styles correctly)
it should be using the color that is defined in `.ef-button-default` (#525859) but it's instead using the keg-components default color

image

daniel112 (changes requested):
looks good. just found 1 minor thing with the evfButton color

tap.js
lancetipton: Yah, metro was throwing because you can't import node core modules in a react native file. This file gets imported into one of the containers. I think it's rootContainer.
I'l add a comment.

src/theme/components/button/evfButton.js
lancetipton: This is fixed. The code to set the correct color was commented out. I added it back it, and it's working properly now

tap.js
lancetipton: Updated with some comments

lancetipton:
Fixed some minor issues brought on in the PR reviews. Also pushed new docker package. It's ready for review again

src/theme/components/button/evfButton.js
lancetipton: This is fixed

daniel112 (approved):
looks good!

PR #58, opened 2020-09-23
Zen 363 loading btn

daniel112:
> Note: depends on https://github.com/simpleviewinc/keg-hub/pull/37

**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-363)

* we need to be able to indicate to the user that we're loading/processing something after a button action
* evf guys mentioned that their app uses font awesome spinner. (FontAwesome 4 spinner to be exact)
* The `Processing` button should have:
   * loading indicator on the left
   * text "Processing" on the right

## Goal

* Update the `EvfButton` for a new prop `isProcessing`
* when `isProcessing` is true, change the content to a custom Processing component 

## Updates

* `src/assets/icons/evf/elements/loading.js`
   * added fontAwesome 4 spinner as SVG Component

* `src/components/button/evfButton.js`
   * Updated for new prop `isProcessing`
   * added `Processing` component which contains the spinner and text

* `src/components/loading/evfLoading.js`
  * added Custom Loading component
  * updated buildStyles so it adjusts the width properly when swapping content to `processing`

* `src/containers/testContainer.js`
  * added examples of the evfButtons

* `src/theme/components/modals/groupBookingModal.js`
   * styling update on the bottom section
## Testing

* Build: `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-363-loading-btn command=test:build`
* Tap: `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-363-loading-btn`
* navigate to http://local.kegdev.xyz/test
* See the new demo section for EvfButtons
* Click on the `Fetch demo` btn to simulate the interaction


src/components/button/evfButton.js
```diff
@@ -1,17 +1,27 @@
...
   const parsedState = { main: custom.parsed }
 
+  // overwrite the min width when trying to display 'processing' content
+  // for cases where the default content is smaller  than the processing content
+  const fitWidth = custom.isProcessing && {
+    minWidth: 'fit-content',
+  }
+
   return theme.get(btnStyles, custom.styles, {
+    main: {
```
mpcarolin, line 22: looks like this could just be `main: fitWidth`

src/assets/icons/evf/elements/loading.js
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react'
```
lancetipton, line 1: Any reason this doesn't use the [SvgIcon from keg-components](https://github.com/simpleviewinc/keg-hub/blob/develop/repos/keg-components/src/assets/icons/svgIcon/svgIcon.js)

src/components/button/evfButton.js
```diff
@@ -29,27 +39,61 @@ const buildStyles = (theme, custom) => {
...
+    </View>
+  )
+}
+
+/**
+ * Processing
+ * @param {object} props
+ * @param {object} props.styles
+ */
+const Processing = ({ styles, size }) => {
```
mpcarolin, line 92: can you add a prop for the processing text here and in `EvfButton`? It could default to `Processing`, but we might need to override that at some point

src/components/loading/evfLoading.js
```diff
@@ -0,0 +1,61 @@
...
+import { EVFIcons } from 'SVIcons'
+import { useTheme } from '@keg-hub/re-theme'
+
+/**
+ * Custom Indicator component using the spinner from FontAwesome 4
+ * @param {object} props
+ * @param {number} props.size
+ * @param {object} props.styles
+ */
+const CustomIndicator = ({ size, styles }) => {
```
mpcarolin, line 16: this is great.

I actually think it could go into `keg-components` as a generic `SpinIndicator` or something. Then you just pass in your icon and now we have a rotating spinner for any icon. 

Just a suggestion, though. We could wait until another use case comes up

mpcarolin (changes requested):
Hey, overall looks really good and passed testing.

It just needs to be rebased with `develop`. I also provided some suggestions in my comments.

src/components/button/evfButton.js
```diff
@@ -1,17 +1,27 @@
...
+ * @param {object} custom - contains {type, styles, isProcessing}
  */
 const buildStyles = (theme, custom) => {
   const btnStyles = theme.get(`button.evfButton.${custom.type}`)
   const parsedState = { main: custom.parsed }
 
+  // overwrite the min width when trying to display 'processing' content
+  // for cases where the default content is smaller  than the processing content
+  const fitWidth = custom.isProcessing && {
+    minWidth: 'fit-content',
```
lancetipton, line 18: The `fit-content` rule does not work for react-native. As a work around on I've been using =>
```js
    flexDirection: 'row',
    alignSelf: 'flex-start',
```
* This works for in the theme, as we can separate between native / web.
* But we can't do that here, when it's inline.
* Can you try using the above two rules instead of ` minWidth: 'fit-content',`
* I'm pretty sure we should be able to swap the above rules in for all instance of `fit-content` in the theme too. I just never tested it.

* For now, if it works, just do it for this since it's defined inline, and we can't use the $native / $web

src/components/loading/evfLoading.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
```
lancetipton, line 1: This is dope. Nice reuse of KegLoading.

src/theme/components/modals/groupBookingModal.js
```diff
@@ -43,23 +43,33 @@ export const groupBookingModal = {
           main: {
             $xsmall: {
               flexDirection: 'row',
-              justifyContent: 'flex-end',
+              alignSelf: 'flex-end',
               paddingTop: 26,
               paddingBottom: 35,
+              flex: 1,
+              width: 'fit-content',
```
lancetipton, line 50: I mentioned it in an above comment, but `fit-content` only works for web.
You should be able to replace it with =>
```js
    flexDirection: 'row',
    alignSelf: 'flex-start',
```

lancetipton (changes requested):
Looks really good. I like how you did the loading. Just a few things to look at.

src/components/button/evfButton.js
daniel112: fitWidth can be null/undefined.
in which case I don't actually want to set `main` as null/undefined (it causes style issues)

src/components/button/evfButton.js
daniel112: can do

src/components/loading/evfLoading.js
daniel112: i can see that. but the custom Indicator is directly tied to `KegLoading` so its not really that generic. 
and the animation hook is already accessible from `keg-components` so there's really not much other work that a consumer needs to do to use a custom icon.
but i do like the idea of being able to just pass in an Icon to the component and it would use that. someday

src/assets/icons/evf/elements/loading.js
daniel112: I thought about using `SvgIcon` but 
I just wanted to keep it consistent with the way the other svg elements are being loaded.
Since we previously had a custom component `EVFIcons` with its own sets of props. I ended up following that route


maybe that's a cleanup item we can do afterwards. update all the icons in `icons/evf/elements` to use SvgIcon instead

src/theme/components/modals/groupBookingModal.js
daniel112: that doesnt seem to work. but i also already have flexDirection and alignSelf set to some value. 
```Javascript
// original obj
$xsmall: {
              flexDirection: 'row',
              alignSelf: 'flex-end',
              paddingTop: 26,
              paddingBottom: 35,
              flex: 1,
              width: 'fit-content',
            },
```
but also. it doesn't seem like i need the `width: 'fit-content` in here. so i can just remove it from the theme

src/components/button/evfButton.js
daniel112: swapping the `minWidth: 'fit-content` doesnt seem to work. its styled the same as if I didn't pass anything in

src/assets/icons/evf/elements/loading.js
lancetipton: If the plan is to eventually merge them all over to `SvgIcon`, then all new code should just use it. No point in creating extra work for ourselves later. Let's just do it now, and update the others later when we have the time.

src/components/button/evfButton.js
lancetipton: Try also adding `minWidth: "100%"`. This case is different then others where `fit-content` was on the `width` rule. Because it's `minWidth`, Adding the `100%` might do the trick

src/theme/components/modals/groupBookingModal.js
lancetipton: works for me

src/assets/icons/evf/elements/loading.js
daniel112: yeah, u right. will update this

src/components/loading/evfLoading.js
mpcarolin: Sounds good.

But my main point is just that it's a fairly common use case to need an icon to spin indefinitely. So much so that FontAwesome [created a class](https://fontawesome.com/how-to-use/on-the-web/styling/animating-icons) just for that. 

That part could be generic. So rather than just repeat this animation wrapper for every one, I think it would make sense down the line to do something like:

```js
const spinners = {
  foo: Spinner(Icons.Foo),
  bar: Spinner(Icons.Bar),
  ...
}
```

if and only if we have more icons to make into spinners of course

src/components/button/evfButton.js
mpcarolin: ah gotcha

src/components/button/evfButton.js
lancetipton: It the above doesn't work. Just add it to the theme in `theme/components/button/evfButton.js` after the `primary` key
Something like this in the evfButton theme file => 
```js
primary: {
  ...
},
processing: {
  main: {
    $web: { minWidth: 'fit-content },
  }
}

```

Then in this file you can reference it in the theme.get call on line 27 =>
```js
  return theme.get(
     btnStyles,
     custom.styles,
      theme?.button?.evfButton?.processing,
     // ...other codes
  }

```

lancetipton (changes requested):
Just one comment. The rest looks good

src/components/button/evfButton.js
daniel112: I moved this to the evfButton theme instead so we can separate by $web and $native

src/components/loading/evfLoading.js
daniel112: gotcha. yeah i dig it

src/theme/components/button/evfButton.js
```diff
@@ -85,12 +88,43 @@ const buttonStateStyles = backgroundColor => {
       },
       content: defaultButtonStyles.content,
     },
+    disabled: {
```
lancetipton, line 91: So looking at this file all these states could be cleaned up like this => 
```js
const buildButtonState = stateStyles => deepMerge({
  main: {
    $all: {
      flex: 1,
      justifyContent: 'center',
      borderRadius: 0,
      padding: 0,
    },
    $web: {
      boxShadow: '0px 1px 1px rgba(0, 0, 0, 0.15)',
    },
  },
  content: defaultTextStyle,
}, stateStyles)


const buttonStateStyles = backgroundColor => {
  return {
    default: buildButtonState({ main: { $all: { backgroundColor }} }),
    active: buildButtonState({ main: { $all: { backgroundColor, opacity: 0.4, } } }),
    hover: buildButtonState({ main: { $all: { backgroundColor, opacity: 0.8, } } }),
    disabled: buildButtonState({ main: { $all: { backgroundColor, opacity: 0.6, } } }),
  }
}
```
this way you only have to define the main button state styles once, and you're not repeating your self

lancetipton (approved):
Looks good, just one small suggestion, but it's not blocking

mpcarolin (approved):
🚀

src/theme/components/button/evfButton.js
daniel112: updated

PR #59, opened 2020-09-25
Keg sessions@0.2.0

lancetipton:
## Context
* The keg-sessions bundle is larger then it needs to be
* There's a number of things we can do to improve the bundle size

## Goal
* Improve the keg-sessions packages by reducing the bundle size
  * I got bundle size down to around 214kb - This is more then 2x smaller then it was before

## Updates
* `.gitignore`
  * Ignores the fonts folder
    * At some point I'm going to update how the fonts are loaded. For local development I want to store the fonts
    * This way we don't have to keep calling google for them
* `configs/rollup.config.js`
  * Added `terser` and `analyze` roll-up plugin
    * terser - Minifies the build. Cuts out about 200kb
    * analyze - Give an overview of what's included in the bundle by size
* `package.json`
  * Updated the version to be `0.2.0`
    * There should be no breaking changes
    * All should be backwards compatible
  * Updated keg-core version to `6.0.0`
  * Removed `moment.js`
    * This is a huge package that was adding extra bloat to the bundle size
    * I replaced it with `date-fns.js`
* `src/components/button/evfButton.js`
  * Cleaned up how the dynamic styles are applied to buttons
  * Was causing issues when no stylesheet exists on the Dom
* `src/components/dates/dayToggle.js` && `src/utils/dateTime/getTimeFromDate.js`
  * Updated date parsing to use `date-fns.js` instead of `moment.js`
  * Fixed `...` issue when the Month name is too long
* `src/components/modals/presenterDetails.js`
  * Updated placeholder image to use a url instead of being converted to a data:base64 string
  * This reduced the bundle size significantly
* `src/hooks/useParsedStyle.js`
  * Fixed bug with classList throwing an error when it didn't exist
* `src/theme/components/dates/dayToggle.js`
  * Fixed an issue with styling the day toggle arrows
* `src/theme/components/sessions.js`
  * Updated how the filter button styles defined the location
  * This makes it much more consistent
  * Cleaned up to always return a default classList object
  * Added condition to only update the fonts and colors when parse styles are returned
* `src/utils/models/sessions/buildHourSessionsMap.js`
  * The date code here was doing the same thing as the `getTimeFromDate` helper method
  * So I updated it to call that method instead of duplicating the code

## Testing

**Consumer**
* Run the consumer: `keg consumer pack run docker.pkg.github.com/simpleviewinc/keg-packages/keg-test-consumer:reducer-sessions-bundle-size`
  * This includes a bundle of all the changes from this pull request
* Navigate to http://local.kegdev.xyz
  * Ensure everything renders as expected
  * Change the browser window size
    * Ensure the filter button location is consistent
    * Ensure the filter button never overlaps the day toggle component
    * Ensure the day toggle component never switches to `...`
    * At a width of `750` the Month should switch to a `3 character version`
* Click the toggle modal demo button
  * Click the second presenter button
  * Ensure the placeholder image still loads

**Tap-Eventsforce**
* Run the tap: `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:keg-sessions-0.2.0`
  * Do all the same tests from the consumer for the tap
  * Ensure it all renders properly

### DIFF
--- a/configs/rollup.config.js
+++ b/configs/rollup.config.js
@@ -1,3 +1,4 @@
+import analyze from 'rollup-plugin-analyzer'
 import babel from '@rollup/plugin-babel'
 import json from '@rollup/plugin-json'
 import resolve from '@rollup/plugin-node-resolve'
@@ -17,6 +18,7 @@ const path = require('path')
 const tapPath = require('app-root-path').path
 const corePath = path.join(`${tapPath}`, `node_modules/keg-core`)
 const isProd = process.env.NODE_ENV === 'production'
+const { ANALYZE } = process.env
 
 const peerExternals = [ 'react', 'react-dom' ]
 const mainExternals = [
@@ -156,5 +158,6 @@ export default {
     cleanup(),
     isProd && terser(),
     bundleSize(),
+    ANALYZE && analyze()
   ],
 }
--- a/configs/rollup.config.js
+++ b/configs/rollup.config.js
@@ -11,10 +11,12 @@ import sourcemaps from 'rollup-plugin-sourcemaps'
 import bundleSize from 'rollup-plugin-bundle-size'
 import generatePackageJson from './plugins/generatePackageJson'
 import image from '@rollup/plugin-image'
+import { terser } from "rollup-plugin-terser"
 
 const path = require('path')
 const tapPath = require('app-root-path').path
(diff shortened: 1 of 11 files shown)


mpcarolin (approved):
Looks good 👍

PR #60, opened 2020-09-28
Zen 275 set session selected

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-275)

## Context

* a session item can be 'booked' by hitting the 'select' button
* if single user, update reducer to isSelected
  * call sessionBookingRequest - to book right away
* if multi-user, call the open modal, to open the group modal selection

## Goal

* Create the action `setSessionSelected`
* Setup the `BookingButton` component in the gridItem so we can easily test this action
   * NOTE: the bookingButton is not complete. it will be fully handled on 
        * https://jira.simpleviewtools.com/browse/ZEN-371
        * https://jira.simpleviewtools.com/browse/ZEN-268
* Fix `sessionLink` color on web (should have primary)

## Updates

* `src/actions/session/setSessionSelected.js`
   * Either books the current session or opens up group booking modal based on the attendees count

* `src/components/button/bookingButton.js`
   * initial booking button that will show up in the grid item

* `src/components/grid/gridTileContent.js`
   * added the bookingButton component

   * ~`colors.color4` was undefined for some reason so the `SessionLink` text color was not getting the right primary color~
       * ~so i added the defaultColors.primary for the case that colors.color4 is undefined or null~
  * updated the setColors to use the default colors if the passed in setColors is undefined/null
## Testing

* `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-275-set-session-selected command=web`
* navigate to http://local.kegdev.xyz/
* Click on `SELECT` button on any grid item
* verify that you see group booking modal appears
* Close the modal
* Change the JSON data > `attendees` array to only contain 1 attendee
    * image

* hit `Apply`
* Click on `SELECT` button on any grid item
* you should NOT see a modal appear, but instead see a console.log

*  Update any `session.capacity.remainingPlaces` to 0 > hit 'Apply'
   * image

* see that the corresponding session item does not have a booking button
    * image


src/actions/session/setSessionSelected.js
```diff
@@ -0,0 +1,36 @@
...
+import { sessionBookingRequest } from 'SVActions/session/booking/sessionBookingRequest'
+
+const { MODAL_TYPES } = Values
+
+/**
+ * Either books the current session or opens up group booking modal
+ * based on the # of attendees available
+ * @param {import('SVModels/session').Session} session
+ */
+export const setSessionSelected = session => {
```
mpcarolin, line 14: I think this could be renamed to something else (`selectSession`?),  because `setSessionSelected` sounds like we are storing the selected session or its id in the items store, but really this is either opening a modal or making a booking request.

src/components/button/bookingButton.js
```diff
@@ -0,0 +1,34 @@
...
+    return null
+
+  // we want to show the SELECT btn if:
+  // - waiting list is available
+  // - remainingPlaces > 0
+  return (
+    <EvfButton
+      type={'primary'}
+      styles={styles}
+      onClick={() => setSessionSelected(session)}
```
mpcarolin, line 30: should put this in a `useCallback`

src/actions/session/setSessionSelected.js
```diff
@@ -0,0 +1,36 @@
...
+  if (!session) return
```
lancetipton, line 15: In `non-production` environments we should log a warning here. That way it doesn't just fail silently
* Might be good to add a log helper to first check the current NODE_ENV, and log a warning 

Could look something like this =>
```js
/**
 * Helper to log messages when not in production
 * @param {string} type - Type of log message. Must be one of log, info, dir, warn, error
 * @param {Array} toLog - Data to be logged
 */
export const devLog = (type, ...toLog) => (
  process.env.NODE_ENV !== 'production' &&
    console[type] &&
    console[type](...toLog)
)
```

Then you can call it like this => 
```js
devLog(`warn`, `Action "setSessionSelected" requires a valid session. Instead received`, session)
```

You could also add helpers like this => 
```js
devLog.log = (...toLog) => devLog(`log`, ...toLog)
devLog.trace = (...toLog) => devLog(`trace`, ...toLog)
devLog.debug = (...toLog) => devLog(`debug`, ...toLog)
devLog.info = (...toLog) => devLog(`info`, ...toLog)
devLog.dir = (...toLog) => devLog(`dir`, ...toLog)
devLog.warn = (...toLog) => devLog(`warn`, ...toLog)
devLog.error = (...toLog) => devLog(`error`, ...toLog)
```

Then you could call like this => 
```js
devLog.warn(`Action "setSessionSelected" requires a valid session. Instead received`, session)
```

If you do this, replace any other log warnings that already exist, like the one in `actions/modals/removeModal.js` on line 35

src/actions/session/setSessionSelected.js
```diff
@@ -0,0 +1,36 @@
...
+
+  const { items } = getStore()?.getState()
+  const attendeesCp = items && Array.from(items.attendees)
```
mpcarolin, line 18: what does the `Cp` mean?

src/actions/session/setSessionSelected.js
mpcarolin: I am also storing the selected session for the group booking modal. I'll probably move that here.

mpcarolin: what is the difference between the selected session and `activeSession` in the store?

mpcarolin (commented):
Looking good and passed testing. I have a couple of questions in the comments.

src/theme/components/grid/gridItem.js
```diff
@@ -6,7 +6,7 @@ export const gridItem = {
   main: {
     $xsmall: {
       backgroundColor: colors.white,
-      height: 200,
+      height: 'fit-content',
```
lancetipton, line 9: This doesn't work on mobile. Try switching it to 
```js
    flexDirection: 'row',
    alignSelf: 'flex-start',
```
If it doesn't look good with that, then use the `$web` and `$native` to separate the styles

src/actions/session/setSessionSelected.js
daniel112: copy

src/theme/components/grid/gridRowContent.js
```diff
@@ -8,6 +8,8 @@ export const gridRowContent = {
     main: {
       flex: 1,
       flexDirection: 'column',
+      paddingBottom: 10,
```
lancetipton, line 11: These could just be
```js
pB: 10,
pL: 10
```
Or you could use the padding helper from keg-components
```js
import { kegComponentsTheme as theme } from 'SVTheme/kegComponentsTheme'

  column2: {
    main: {
      flex: 1,
      flexDirection: 'column',
      ...theme.padding(10, ['bottom', 'left']),
    }

```

lancetipton (changes requested):
Looks good. Only a few minor things. Some are just suggestions. You don't have to do it.

src/actions/session/setSessionSelected.js
daniel112: `activeSession` is not used atm.
i think the idea behind it was to determine which session item the user clicked on for mobile expand. but we're not going down that route anymore i dont think

src/components/button/bookingButton.js
daniel112: updated

src/actions/session/setSessionSelected.js
daniel112: il update the name

src/theme/components/grid/gridRowContent.js
daniel112: updated to use shortCut

src/theme/components/grid/gridItem.js
daniel112: updated to platform specific styles

src/actions/session/setSessionSelected.js
daniel112: added this helper in the utils
updated `selectSession`, `removeModal`, `formatPrice` and `widthOf`

daniel112:
updated from feedback

src/components/button/bookingButton.js
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * @todo - to be completed on
+ *       - https://jira.simpleviewtools.com/browse/ZEN-371
+ *       - https://jira.simpleviewtools.com/browse/ZEN-268
+ * @param {object} props
+ * @param {object} props.styles
+ * @param {import('SVModels/session').Session} props.session
+ */
+export const BookingButton = ({ styles, session }) => {
+  if (!session) return null
```
lancetipton, line 13: This could be joined with the condition on line 17.

lancetipton (approved):
Only one small thing, and it's not really part of this ticket, so no big deal

PR #61, opened 2020-09-30
Zen 285 apply filter

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-285)

## Context

* the filter modal has filter items that users can select 1 or more of
* the user can then press the `apply` button that applies and saves the filter item selected
* see the comments on the ticket for detailed information on how the filtering should work

## Goal

* The filter modal should be able to filter the sessions

## Updates

* `src/actions/session/filters/applySessionFilters.js`
   * filter by states first, then filter by labels
   * updates the session agenda store afterwards

* `src/actions/session/filters/cancelSelectedFilters.js`
   * for when user closes the filter modal without hitting `apply`
   * undoes any changes that's not been set to be the ACTIVE filter

   * extracted the `mapAgendaSessions` to its own action so it can be re-used

* `src/actions/session/setAgendaSessions.js`
   * extracted from mapSessionInterface

   * added the apply action and cancel action
   * minor cleanup on the label toggling logic
   * filter out the booking states that we dont want used as a filter

* `src/constants/values.js`
   * added more booking states

* `src/utils/models/sessions/getBookingState.js`
   * helper method to get the booking state for a given session

### Unit test
* Run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-285-apply-filter command=test`

### With the default Test Data
* run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-285-apply-filter`

#### Filter by label tags
- Open Filter modal > Filter by `Important` tag
- there should be 1 session on day 1 and 2. 2 sessions on day 3
- update the filter to be `Important` and `Breaks`
- there should now be 2 sessions on day 2 (no changes on day 1 and 3)
    - image

#### Filter by booking states. see [link](https://jira.simpleviewtools.com/browse/ZEN-285?focusedCommentId=58811&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-58811) for detailed context on each one

> If the filtering is applied by the booking state filters “Available”, “Selected”, “Waiting List” or “On Waiting List”, then when a session with `allowBooking` == false should NOT be displayed in the results
  - change to day 1
  - filter by 'Big Data' only
  - see that only 1 session appears on day 1
  - now add any of the booking state filters on top of 'Big Data'
  - see that no session appears on day 1
     - this is because `allowBooking` is false on that one session

### Verify the following `booking state` logic applies the filtering correctly: (modifying the JSON data accordingly)
1. if `session.allowBooking` === false. filtering by any booking state filter should NOT display this session
2. **Available**:  if the `session.identifier` DNE on ANY `attendee.waitingListSessions` or `attendee.bookedSessions` AND (`session.capacity.isUnlimited` == true OR `session.capacity.remainingPlaces > 1`)
3. **Selected**:  Any session where the `session.identifier` is included in the `attendee.bookedSessions` array for any attendee
4.  **Waiting list**: `session.identifier` DNE on  `attendee.bookedSessions` or `attendee.waitingListSessions` AND `session.capacity.isUnlimited == false` + `session.capacity.remainingPlaces == 0` + `session.capacity.remainingPlaces.isWaitingListAvailable == true` 
5. **On waiting list**: Any session where the `session.identifier` is included in the `attendee.waitingListSessions` array for any attendee
6.  If a session.identifier exists in both attendee.waitingListSessions && attendee.bookedSessions, **On waiting List** should take priority for the state. [reference](https://jira.simpleviewtools.com/browse/ZEN-285?focusedCommentId=58981&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-58981)


src/actions/session/mapSessionInterface.js
```diff
@@ -71,7 +44,7 @@ export const mapSessionInterface = props => {
         // while mapping sessions, also map for 'activeSessions'
         // activeSessions are sessions sorted by day number
         if (key === CATEGORIES.SESSIONS)
-          mapAgendaSessions(value, props.agendaDays)
+          setAgendaSessions(value, props.agendaDays)
```
lancetipton, line 47: This is great. It cleans this file up a lot. I think there's more that can be done too. Here's what I came up with this this file =>
```js
import { dispatch } from 'SVStore'
import { ActionTypes, Values } from 'SVConstants'
import { mapObj, snakeCase } from '@keg-hub/jsutils'
import { addModal } from 'SVActions/modals'
import { Modal } from 'SVModels/modal'
import { setAgendaSessions } from 'SVActions/session/setAgendaSessions'

const { CATEGORIES, SUB_CATEGORIES } = Values

/**
 * defines which sessionAgendaProps need to be mapped to subcategories
 */
const subCatMap = {
  settings: SUB_CATEGORIES.AGENDA_SETTINGS,
}

/**
 * If alert prop is valid, add a new modal item for alert
 * @param {<import('SVModels/alert').Alert)>} alert
 */
const checkAlert = alert => {
  alert?.title &&
    alert?.message &&
    addModal(new Modal({ type: CATEGORIES.ALERT.toLowerCase(), data: alert }))
}

const getDispatchPayload = (key, value) => {
  return !subCatMap[key]
    // by default, we use set items, so that if the component is mounted/remounted, data won't be duplicated
    ? {
        type: ActionTypes.SET_ITEMS,
        payload: { category: key, items: value }
      }
    // subcategories are upsert-merged, rather than set, since they
    // might need to be joined with data that was loaded from localStorage,
    // e.g. agendaSettings.activeDayNumber
    : {
        type: ActionTypes.UPSERT_ITEM,
        payload: { category: key, item: value, key: subCatMap[key] }
      }
}

/**
 * push the sessionAgendaProps items to our local state
 * @param {import('SVModels/sessionAgendaProps').SessionAgendaProps} props
 */
export const mapSessionInterface = props => {
  // loop through each key and dispatch accordingly
  props &&
    mapObj(props, (key, value) => {
      // ensure key exists in the items store first
      if(key !== CATEGORIES[snakeCase(key).toUpperCase()]) return

      // while mapping sessions, also map for 'activeSessions'
      // activeSessions are sessions sorted by day number
      key === CATEGORIES.SESSIONS &&
        setAgendaSessions(value, props.agendaDays)
      
      // Check for the alert prop, and call the checkAlert when it exists
      // Otherwise just dispatch the payload based on the key and value
      key === CATEGORIES.ALERT
        ? checkAlert(props.alert)
        : dispatch(getDispatchPayload(key, value))
    })
}

```

src/components/modals/filter.js
```diff
@@ -103,13 +115,18 @@ const LabelButtons = ({ styles, labels }) => {
...
-  const isFilterEmpty = !filters.activeFilters.length && !selectedCount
+  const isFilterEmpty = !selectedCount
 
   return labels.map(label => {
-    // Check selectedCount before doing the loop. If none are selected, we same a few cpu cycles
-    const isLabelOn =
-      selectedCount &&
-      filters.selectedFilters.some(item => item.identifier === label.identifier)
+    // Check selectedCount before doing the loop. If none are selected, we save a few cpu cycles
+    const isLabelOn = useMemo(() => {
```
lancetipton, line 122: This is great, but I would pull out, and put it in hook outside the loop. Like this => 
```js

// Above the LabelButtons component
const useLabelOn = (selectedCount, selectedFilters, item, label) => {
  // Check selectedCount before doing the loop. If none are selected, we save a few cpu cycles
  return useMemo(() => {
    return (
      selectedCount &&
      selectedFilters.some(
        item => item.identifier === label.identifier
      )
    )
  }, [ selectedFilters, selectedCount, label, item ])

}

 // Replace line 122 with this
const isLabelOn = useLabelOn(selectedCount, filters.selectedFilters, item, label)

``` 

It makes the code a lot easier to read and reason about.

src/components/sessions.js
```diff
@@ -144,6 +145,7 @@ export const Sessions = props => {
 
   useEffect(() => {
     mapSessionInterface(sessionAgendaProps)
+    applySessionFilters()
```
lancetipton, line 148: This looks like it could be a race condition. **I did a number of tests, but could not make it fail, so I'm just marking this incase the issue comes up at a later time.**

 * If the code in `mapSessionInterface` does not finish before the code in `applySessionFilters` is run, then the `applySessionFilters` might not get the most up-to-date data that it needs to filter the items
 * Also internally both actions call `setAgendaSessions`. So if `applySessionFilters` calls `setAgendaSessions` before `mapSessionInterface`
  * Then the items from `mapSessionInterface` will be what's in the store, and not the items from `applySessionFilters`
  * This could happen if `mapSessionInterface` is ever updated to be an **asynchronous call**

lancetipton (changes requested):
Overall looks really good. There's a few small things to look at, but nothing major. 
There's one item I just left a note on, as I think it could be an issue, but I could not get it to fail; so for now, don't worry about it

src/components/sessions.js
daniel112: yeah that makes sense if either of those functions are async, but since neither are, i dont think the race condition could occur here.
I agree that if we ever update this to be async we would have to address this

src/actions/session/filters/applySessionFilters.js
```diff
@@ -1,15 +1,82 @@
...
  */
 export const applySessionFilters = () => {
-  // 1. sets the current selectedFilters to activeFilters
-
-  // for now just print out the currently selected filters
   const { items } = getStore()?.getState()
   const selectedFilters = items?.filters?.selectedFilters || []
-  console.log('current selected filters: ')
-  console.log(...selectedFilters)
+  const sessions = items?.sessions
```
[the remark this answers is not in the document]
daniel112: well on a case where:
1. user opens filter modal > applies some filter to the session > hit apply
    - sessions are now filtered accordingly
2. user opens the filter modal > unselect all the filters (essentially clearing) > hit apply
    - sessions should reset, so I would still need to call 
            - `setAgendaSessions` but with the original non filtered sessions
            - dispatch for `ACTIVE_FILTERS` so it no longer has any values in the array

so i could update it to something like
```
let filteredSessions = sessions
  if (selectedFilters.length > 0) {
    // filter by states
    let filteredSessions = applyStateFilters(selectedFilters, sessions)

    // filter by labels
    filteredSessions = applyLabelFilters(selectedFilters, filteredSessions)
  }
   // update agenda sessions store
   setAgendaSessions(filteredSessions, agendaDays)

  // set the current selectedFilters to activeFilters
  dispatch({
    type: ActionTypes.SET_ITEM,
    payload: {
      category: CATEGORIES.FILTERS,
      key: SUB_CATEGORIES.ACTIVE_FILTERS,
      item: selectedFilters,
    },
  })

```

src/actions/session/filters/applySessionFilters.js
```diff
@@ -1,15 +1,82 @@
...
+}
+
+/**
+ * Filter by booking states
+ * @param {Array.<import('SVModels/label').Label>} selectedFilters
+ * @param {Array.<import('SVModels/session').Session>} sessions
+ *
+ * @returns {Array.<import('SVModels/session').Session>} - list filtered by button state or original list
+ */
+const applyStateFilters = (selectedFilters, sessions) => {
```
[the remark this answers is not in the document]
daniel112: updated to `sessionsFromStateFilters `

src/actions/session/filters/applySessionFilters.js
```diff
@@ -1,15 +1,82 @@
...
+}
+
+/**
+ * Filter by labels
+ * @param {Array.<import('SVModels/label').Label>} labels
+ * @param {Array.<import('SVModels/session').Session>} sessions
+ *
+ * @returns {Array.<import('SVModels/session').Session>}
+ */
+const applyLabelFilters = (labels, sessions) => {
```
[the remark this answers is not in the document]
daniel112: updated to `sessionsFromLabelFilters `

src/actions/session/filters/applySessionFilters.js
```diff
@@ -1,15 +1,82 @@
...
+    type: ActionTypes.SET_ITEM,
+    payload: {
+      category: CATEGORIES.FILTERS,
+      key: SUB_CATEGORIES.ACTIVE_FILTERS,
+      item: selectedFilters,
+    },
+  })
+}
+
+/**
```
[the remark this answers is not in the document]
daniel112: updated

src/components/modals/filter.js
daniel112: will do. updated

src/actions/session/mapSessionInterface.js
daniel112: works for me. updated

src/actions/session/filters/applySessionFilters.js
```diff
@@ -15,13 +16,19 @@ export const applySessionFilters = () => {
   const sessions = items?.sessions
   const agendaDays = items?.agendaDays
 
-  // 1. filter by states
-  let filteredSessions = applyStateFilters(selectedFilters, sessions)
+  let filteredSessions = sessions
+  if (selectedFilters.length > 0) {
+    // filter by states
+    filteredSessions = sessionsFromStateFilters(selectedFilters, sessions)
```
lancetipton, line 22: Nothing wrong with this, but I would do it like this instead
```js
  const filteredSessions = selectedFilters.length > 0
    ? sessionsFromLabelFilters(
        selectedFilters,
        sessionsFromStateFilters(selectedFilters, sessions)
      )
    : sessions
```
One reason is because `let`'s  take more memory, so I try to use constants as much as possible.
But the main reason is, I just don't like `if ( some condition ){ ...do the thing }` 
  * It always seems so verbose to me
  * But again nothing wrong with what you did

src/actions/session/filters/applySessionFilters.js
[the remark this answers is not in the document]
lancetipton: Yah, that makes sense. Good point

src/actions/session/mapSessionInterface.js
```diff
@@ -25,30 +25,24 @@ const checkAlert = alert => {
 }
 
 /**
- *  map AgendaSessions using sessions and agenda day numbers
- * @param {Array<import('SVModels/session').Session>} sessions
- * @param {Array} agendaDays
+ *
```
lancetipton, line 28: Can you add a description here

lancetipton (changes requested):
Code looks good. Saw 2 very small things, and one of them does not require an update if you don't want to.

src/actions/session/filters/applySessionFilters.js
daniel112: got it. updated

src/actions/session/mapSessionInterface.js
daniel112: yup. updated

lancetipton (approved):
Looks good, thanks

PR #62, opened 2020-09-30
fixed date parsing for safari

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-386)

## Context

* With the latest update on `getTimeFromDate` to use the built in `Date` constructor. It introduced a bug on safari where it does not properly create a `Date` object if the formatting is "incorrect"
* we're passing in a string to this helper function and it is formatted like: `2020-08-03 13:00:00`
   * safari does not support parsing the dashes `-`. but it does support `/`
   * so just replace the dashes with slashes if the parsing fails
## Goal

* Update `getTimeFromDate` to format the date to an acceptable format for browser that complains about it

## Updates

* `src/utils/dateTime/getTimeFromDate.js`
   * parse the date string for safari / any other browser that doesn't support `-`

## Testing

* run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:safari-date-parse-bug`
* run the app on safari > verify that it doesn't throw an error/crash
* run on the other browsers to make sure it still works the same/doesn't crash

### DIFF
--- a/src/utils/dateTime/getTimeFromDate.js
+++ b/src/utils/dateTime/getTimeFromDate.js
@@ -7,7 +7,7 @@ import { format } from 'date-fns'
  * @returns {string}
  */
 export const getTimeFromDate = (date, military = true) => {
-  let parsed = Date.parse(new Date(date.replace(/-/g, '/')))
+  let parsed = Date.parse(date)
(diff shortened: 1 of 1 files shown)


lancetipton (approved):
I like it, super clean. Works well

mpcarolin (approved):
Looks good 👍

PR #64, opened 2020-10-01
Sort agendaSessions by timeblock

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-390)

## Context

* I noticed that our agenda sessions grid isn't actually sorting the sessions by time block. it just so happens that the test data we've been using is in the correct order

## Goal

* have the `AGENDA_SESSIONS` store actually be sorted by timeblock

## Updates

- updated unit test for buildHourSessionsMap
- updated buildHrSessionsMap to return an array instead of an object so we can sort it by time block

## Testing

* run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:sort-by-hour-block`
* navigate to http://local.kegdev.xyz/
* switch to Day 1
* modify the JSON data to add a new session on day 1, and set the start time > 9 am  && < 1 pm
   * something like this
      image

* hit apply
* verify that the new session block gets added in between `9:00` block and `13:00`

### unit test
* run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:sort-by-hour-block command=test`

### DIFF
--- a/src/utils/models/sessions/buildHourSessionsMap.js
+++ b/src/utils/models/sessions/buildHourSessionsMap.js
@@ -1,31 +1,41 @@
 import { getTimeFromDate } from '../../dateTime/getTimeFromDate'
 import { sortSessions } from './sortSessions'
-import { reduceObj } from '@keg-hub/jsutils'
 
 /**
  * Filters the sessions given a day number
  * @param {Array<import('SVModels/session').Session>} sessions
  * @param {number} dayNumber
  * @param {boolean=} asc - order ascending by name
- * @returns {object} - where key is the hour, value is the array of sessions
- *                   - ex: {6:[], 7:[sessionA, sessionB, etc], 8:[sessionC, etc]}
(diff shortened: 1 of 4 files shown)


src/utils/models/sessions/buildHourSessionsMap.js
```diff
@@ -1,31 +1,41 @@
...
+            (items[mapIndex].sessions || []).concat(session),
+            asc
+          ))
+        : items.push({ timeBlock, sessions: [session] })
+
+      return items
+    }, [])
+    .sort(
+      (itemA, itemB) =>
+        new Date(`1970/01/01 ${itemA.timeBlock}`) -
```
lancetipton, line 38: This works, but feels like there's a better way to do this. At some point this is something we should investigate.

lancetipton (approved):
Looks good. Works for me

PR #70, opened 2020-10-15
fixed bug where booked tickets didnt get flattened

mpcarolin:
## Context
* there was a bug with the mapping of the tickets based on the recent change, the sub tickets weren't getting flattened

## Goal
* fix bug

## Testing
* `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:booked-ticket-mapping-fix`
* ensure that the `/test` page and the group booking modal loads, and verify that you see no warnings in the console about not finding a valid ticket (it would say "Could not find a valid ticket for attendee")


daniel112 (approved):
verified that there are 8 total attendees after the change.
previously only showed 6

PR #74, opened 2020-10-16
Modal Style Updates + Fixed Tests

mpcarolin:
## Context
* modal padding was not consistent and each modal was redefining its own padding/margins'
* in group booking modal, very long names that were also on waiting list weren't wrapping
* `useBookingLists` tests were failing

## Goal

* Make all modal bodies use the same padding, to be consistent and less brittle
* Make long waiting-list names wrap on small screen widths
* Fix `useBookingLists` tests

## Updates

  * updates to get it wrapping
  * updated to wrap modal children in a view that can define consistent padding in a single place
* `src/hooks/booking/__tests__/useBookingLists.js`
  * fixed tests by mocking out `useMemo`, also added a `jest.resetAllMocks()` call after all tests run
* `**/**.js`
  * updated all the other modals and themes to remove inconsistent margins and padding, except where it needed additional padding

## Testing

* `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:consistent-modal-padding`
* for each modal (alert, group booking, session details, filter, presenter):
  * verify everything looks like you would expect
  * verify the content of the modal lines up the the modal header's title
* run `yarn test`, verify all tests pass
  * EXCEPT `useInterval`. Something is still breaking this test (`useRef` is blocking and preventing the test from completing before the timeout), but it precedes my group booking changes so I will investigate this in a separate PR

### DIFF
--- a/src/theme/components/modals/filterModal.js
+++ b/src/theme/components/modals/filterModal.js
@@ -123,9 +123,6 @@ export const filterModal = {
     },
     body: {
       main: {
-        paddingLeft: 46,
-        paddingRight: 26,
-        paddingTop: 17,
         paddingBottom: 26,
       },
       topSection: {
@@ -169,7 +166,7 @@ export const filterModal = {
         applyButton: {
           main: {
             minHeight: 45,
-            mL: 15
+            mL: 15,
           },
         },
         clearButton: clearButtonDefault,
--- a/src/theme/components/modals/filterModal.js
+++ b/src/theme/components/modals/filterModal.js
@@ -122,9 +122,7 @@ export const filterModal = {
       },
     },
     body: {
-      main: {
-        paddingBottom: 26,
-      },
+      main: {},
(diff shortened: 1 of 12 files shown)


lancetipton (approved):
Looks good to me. Modals look good on all sizes and content is aligned.

src/theme/components/modals/filterModal.js
```diff
@@ -123,9 +123,6 @@ export const filterModal = {
     },
     body: {
       main: {
-        paddingLeft: 46,
-        paddingRight: 26,
-        paddingTop: 17,
         paddingBottom: 26,
```
daniel112, line 126: i dont think you need this bottom padding anymore. 
image

daniel112 (changes requested):
just 1 minor thing.
also dont forget to change the base merge to `develop`

PR #76, opened 2020-10-20
Zen 369 clear all

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-369)

## Context

* After sessions are filtered, there should be a way to clear them from the sessions list. Otherwise the user will have to open the filter modal and hit clear filter > apply
* this is a small nice-to-have improvement to save the user a few clicks to clear their filters

## Goal

* Add a `Clear` button on the session agenda next to the `Filter` button
* This clear button should only appear if there are any active filteres
* clicking this button will clear all active filters and apply it to the agenda sessions

## Updates

* `src/components/dates/dayToggle.js`
   * added specific width for the dayToggle string to transform earlier to the mobile view version

   * changed the name to `clear all` because that's what it says on the x5 doc

* `src/components/sessions.js`
   * added the `clear` button next to the `filter` button
   * added some specific styling based on dim.width
   * Clear button will NOT show up once the dim.width is 520px or less. got no space
   * on click: clears the active filters and apply them

* `src/containers/gridContainer.js`
   * removing unecessary code

* `src/theme/components/grid/gridTileContent.js`
   * small fix on grid item wrapping

* `src/theme/containers/gridContainer.js`
   * small fix on the timeblock header wrapping
## Testing

* Run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-369-clear-all`
* Verify you don't see the `Clear` button next to the `Filter` button on app start
* Apply some filtering
* Verify that you now see the `Clear` button appear next to the `Filter` btn
* Click the `Clear` btn
* Verify that it clears out your filter and updates the session agenda items
* Open the Filter modal
* verify that none of the filter label is selected (should all be lit up)

### Mobile web
* change to mobile view width
* apply filters
* verify that `Clear` btn does not show up

### DIFF
--- a/src/components/sessions.js
+++ b/src/components/sessions.js
@@ -30,12 +30,15 @@ const FilterButton = ({ onClick, styles }) => {
 
   // use filter icon when below 650px width
   return dim.width <= 650 ? (
-    <View className={'ef-sessions-filter-button'}>
+    <View
+      className={'ef-sessions-filter-button'}
+      style={styles?.filterIcon?.main}
+    >
       <EVFIcons.Filter
         className={'ef-sessions-filter-button'}
-        style={styles?.filterIcon}
+        style={styles?.filterIcon?.icon}
         onPress={onClick}
-        color={styles?.filterIcon?.color}
+        color={styles?.filterIcon?.icon?.color}
       />
     </View>
   ) : (
@@ -98,6 +101,7 @@ const ItemHeaderRight = ({ styles, onClick }) => {
   return (
     <View style={styles?.main}>
       <Button
+        themePath='button.text.default'
         styles={styles?.clearAll}
         content={'Clear All'}
       />
--- a/src/components/sessions.js
+++ b/src/components/sessions.js
@@ -1,5 +1,5 @@
 import React, { useEffect, useCallback, useMemo } from 'react'
-import { useTheme, useDimensions } from '@keg-hub/re-theme'
+import { useTheme, useDimensions, useStylesCallback } from '@keg-hub/re-theme'
 import { View, ItemHeader, Button, ScrollView } from '@keg-hub/keg-components'
 import { RenderModals } from 'SVComponents/modals/renderModals'
 import { mapSessionInterface } from 'SVActions/session/mapSessionInterface'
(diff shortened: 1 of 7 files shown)


src/components/sessions.js
```diff
@@ -17,40 +20,58 @@ import { Values } from 'SVConstants'
...
-const { EVENTS } = Values
+const { EVENTS, CATEGORIES, SUB_CATEGORIES } = Values
+
 /**
  * FilterButton
  * Renders either an Icon or a text button based on current screen dimension
  * @param {object} props
  * @param {object} props.styles
  * @param {Function} props.onClick
+ * @param {boolean} props.smallEnough - whether the dimension width is small enough
```
mpcarolin, line 31: I think a clearer and more general prop name would be `showIcon`, rather than `smallEnough`, which just leads me to ask "small enough for what?"

src/components/sessions.js
```diff
@@ -87,15 +108,67 @@ const SessionsHeader = ({ styles, onDayChange, labels }) => {
...
+ * @param {object} props
+ * @param {object} props.styles - theme path: header.content.right.content
+ */
+const ItemHeaderRight = ({ styles, onClick }) => {
+  const dim = useDimensions()
+  const activeFilters = useStoreItems(
+    `${CATEGORIES.FILTERS}.${SUB_CATEGORIES.ACTIVE_FILTERS}`
+  )
+
+  const smallEnough = dim.width <= 720
```
mpcarolin, line 133: see comment above about `smallEnough`

mpcarolin (approved):
Approving this 👍 , though I have a comment about a prop name I hope you can look at.

I also don't know how I feel about the `Clear` being underlined too, but that's minor and subjective and if you think it looks good then that works for me:

image

src/components/sessions.js
daniel112: technically it's being used for more than just to show the icon or not. its to update the padding styling in this context. so im a little hesitant to call it `showIcon` here. but im down for coming up with a diff name if you have one

src/components/sessions.js
daniel112: yeah the name is kinda meh.  i can change it for the prop there to `showIcon` since its used to determine that within the `FilterButton` component.

daniel112: updated to `showIcon`

src/components/sessions.js
daniel112: updated to smallWidth

lancetipton (approved):
Overall looks really good. 
* Only thing that seems off is using `clear` vs `clear filters` || `clear all`
* Not a big deal, just not what I expected.

PR #78, opened 2020-10-22
filter results counter

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-375)

`[semver]: minor`

## Context

* Filter Modal specs has changed since we last sized them
* it should also include a 'results' counter (see attached image)
* it should be styled with the same style as the Group Booking Modal - "Places Remaining" counter

## Goal

* add a 'results' counter in the location as seen in the attached image
* styling should be the same as the counter on group booking modal
* Filtering logic should execute before 'APPLY' is pressed because we need to be able to get the count
* get the count number by filtering on the sessions item state
* if results is 0. the 'APPLY' button should be disabled

## Updates

- `src/components/modals/filter.js`
   * updated apply button to be disabled when results == 0
   * updated the classname of the TopSection to be the correct ones
   * added results counter to the top section
   * performs the filtering after each label select/deselect. but does not apply them to the sessionAgenda until Apply is hit

- `src/theme/components/modals/filterModal.js`
   * styling

- added some unit tests to the filtering utility function
 
## Testing

* run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-375-results-counter`
* open the filter modal
* start selecting the labels
* verify that you see the results counter appear with a number
    * image

* Apply it
* verify that the amount of sessions displayed matches the count number from the filter modal

### Result 0 test
* open the filter modal
* select a label combination that leads to result = 0
* Verify that the `Apply` button is disabled

### Unit Test
* run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-375-results-counter command=test`
* verify all test passes


src/components/modals/filter.js
```diff
@@ -65,22 +68,37 @@ export const Filter = ({ visible, labels }) => {
...
+  ])
+  const hasSelectedFilters = Boolean(filters?.selectedFilters.length)
+
+  // do basic filtering on SELECTED labels so we can get the count
+  const filteredSessions =
+    hasSelectedFilters &&
+    sessionsFromLabelFilters(
+      filters?.selectedFilters,
+      sessionsFromStateFilters(filters?.selectedFilters, sessions)
+    )
```
mpcarolin, line 83: I think you could extract this into a hook, like `useFilteredSessions`

```js
const useFilteredSessions = () => {
  const { filters, sessions } = useStoreItems([
    ({ items }) => pickKeys(items, ['filters']),
    CATEGORIES.FILTERS,
    shallowEqual
    CATEGORIES.SESSIONS,
    )
  ])
  return useMemo(() => {
    const hasSelectedFilters = Boolean(filters?.selectedFilters.length)

    // do basic filtering on SELECTED labels so we can get the count
    return hasSelectedFilters &&
      sessionsFromLabelFilters(
        filters?.selectedFilters,
       sessionsFromStateFilters(filters?.selectedFilters, sessions)
      )
  }, [ filters, sessions ])
}
```

Then it would be memoized, reusable, and it abstract these details away from what `Content` component needs to know.

mpcarolin (changes requested):
Just one code suggestion and one comment on the UI:

When there are 0 session to show, we should show a plural `results`:

image

(fun fact: 0 is grammatically equivalent to "not any", so "not any results" === "zero results")

src/components/modals/filter.js
daniel112: I think I will use the jsutils `memorize` and extract this into its own function.
so i can get me memoized benefit && use it in the apply filter action as well

lancetipton: I like @mpcarolin 's Idea.
The jsutils memorize could work, but then it would work different from what we've been doing for all other logic like this.
  * Which is create a hook for it

You can still extract the logic into its own helper which could then be called in the action, but also create a hook, that can be used within the component.
  * This would keep things consistent with the rest of the app

src/components/modals/filter.js
```diff
@@ -234,6 +285,7 @@ const BottomSection = ({ styles, onButtonPress, hasSelectedFilters }) => {
         />
       ) }
       <EvfButton
+        disabled={disableApply}
```
lancetipton, line 288: This [PR](https://github.com/simpleviewinc/tap-events-force/pull/75) has been merged, so you should be able to rebase the changes, so we can actually test the disabled functionality.

lancetipton (changes requested):
Looks good, just a couple things to look at. Thanks

src/components/modals/filter.js
daniel112: discussed with both Lance and Michael.
will update to use hook

src/components/modals/filter.js
daniel112: rebased. this works now

src/hooks/sessions/useFilteredSessions.js
```diff
@@ -0,0 +1,32 @@
...
+  return useMemo(() => {
+    const hasSelectedFilters = Boolean(filters?.selectedFilters.length)
+
+    // do basic filtering on SELECTED labels so we can get the count
+    return (
+      hasSelectedFilters &&
+      sessionsFromLabelFilters(
+        filters?.selectedFilters,
+        sessionsFromStateFilters(filters?.selectedFilters, sessions)
+      )
```
mpcarolin, line 29: So, I think the JSDoc should indicate this function returns `false` if there are no selected filters. That, or update this to return at least an empty array.

mpcarolin (approved):
Looks good. I have one more suggestion, but it's not blocking for me.

src/hooks/sessions/useFilteredSessions.js
daniel112: updated

src/components/modals/filter.js
lancetipton: yes it does

lancetipton (approved):
Looks good to me.

PR #79, opened 2020-10-26
GridItem Height fix

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-425)

## Context

* Each grid item has a dynamic height and minHeight of 294px (from mockup)
* The grid item(s) in each block container (IF they are on the same row) should have the same height
    * it should not do this
       * image

## Goal

* Update styling on the grid item and gridTileContent so that the height of each item grows accordingly

## Updates
- `src/mocks/eventsforce/testData.js`
    - updated mockdata with more presenters
- `src/components/sessionLink/sessionLink.js`
    - added className to sessionLink, so we can override it for presenter name classname
- `src/theme/components/grid/gridTileContent.js`
    - wrapped presenterNames in a container so it can be dynamically grow
- `src/theme/components/grid/gridItem.js`
    - removed fit-content from gridItem, so each grid item can grow if they need to
- `src/components/grid/gridTileContent.js`
    - updated styling on grid tile content so the button is relative

## Testing

* Run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-425-griditem-height`
* navigate to **Day 2** > Timeblock 11:00 AM.  OR add a bunch of session item to a specific time block
* Verify that all the items on the same row have the same height
     * image

* Update the content of a session in that timeblock.
    * *I Modified the location name*
* verify that the height adjusts accordingly for sessions in the same row
    * image

### Unit test
* `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-425-griditem-height command=test`
* ensure all test passes

### DIFF
--- a/src/components/grid/gridTileContent.js
+++ b/src/components/grid/gridTileContent.js
@@ -115,15 +115,17 @@ const PresenterNames = React.memo(({ session, styles }) => {
 
   return (
     <View style={styles?.main}>
-      { presenters.map(presenter => {
-        return (
-          <PresenterLink
-            key={presenter.identifier}
-            styles={styles?.sessionLink}
-            presenter={presenter}
-          />
-        )
-      }) }
+      <View style={styles?.container}>
+        { presenters.map(presenter => {
+          return (
+            <PresenterLink
+              key={presenter.identifier}
+              styles={styles?.sessionLink}
+              presenter={presenter}
+            />
+          )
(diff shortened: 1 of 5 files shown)


lancetipton (approved):
Nice quick fix. Great work on this.

PR #80, opened 2020-10-27
Zen 424 session link underline

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-424)

## Context

* Going back over the RegX5 session doc here https://docs.google.com/document/d/1bZIMvNhg6fsU5Tf1edLcoBfpu8rR9qQLGf0Kx-DGMrw/edit#
* minor detailed that was missed is that the link should become underlined when hovered

## Goal

* Make session link and presenter link underline on hover

## Updates

* `src/components/sessionLink/sessionLink.js`
  * added useThemeHover hook
  * added the theme ref to the Text component

* `src/theme/components/sessionLink.js`
   * added default and hover state styles

## Testing

* run `keg dpg run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-424-session-link-underline`
* verify that the Session link and presenter link is underlined when you hover them 
* image

### DIFF
--- a/src/theme/components/sessionLink.js
+++ b/src/theme/components/sessionLink.js
@@ -1,5 +1,4 @@
 import { colors } from '../colors'
-import { deepMerge } from '@keg-hub/jsutils'
 
 const mainStyle = {
   $xsmall: {
@@ -41,10 +40,12 @@ export const sessionLink = {
     text: textStyle,
   },
   hover: {
-    text: deepMerge(textStyle, {
+    text: {
       $all: {
-        $xsmall: { txDL: 'underline' },
+        $xsmall: { 
+          txDL: 'underline' 
+        },
       },
-    }),
+    },
   },
 }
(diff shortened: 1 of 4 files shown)


src/theme/components/sessionLink.js
```diff
@@ -1,36 +1,50 @@
...
-        lineHeight: 19,
+}
+
+export const sessionLink = {
+  default: {
+    main: mainStyle,
+    text: textStyle,
+  },
+  hover: {
+    text: deepMerge(textStyle, {
```
lancetipton, line 44: Pretty sure you don't need to do a deepMerge here.
The `default` and `hover` styles should already be merged within the `useThemeHover` hook
Unless you explicitly state that they should not be. [See the code here](https://github.com/simpleviewinc/keg-hub/blob/a99901664c23b217659be6a345fb7e3a2ebfe334/repos/re-theme/src/hooks/hookFactory.js#L115)

lancetipton (approved):
Code looks good. Just one thing to look into.
Going to approve because all code is working as it should
But please check the comment, and validate it before merging.

Thanks

src/theme/components/sessionLink.js
daniel112: yup. just verified. will remove deepmerge

PR #87, opened 2020-11-09
Zen 385 remove deps

lancetipton:
**Ticket**:
* [ZEN-384](https://jira.simpleviewtools.com/browse/ZEN-384)
* [ZEN-385](https://jira.simpleviewtools.com/browse/ZEN-385)

> **IMPORTANT**
> This PR should be tested alongside these PRs
> [Keg-Hub](https://github.com/simpleviewinc/keg-hub/pull/69)
> [Keg-Consumer](https://github.com/simpleviewinc/keg-test-consumer/pull/6)

## Context
* Currently, component prop-types are being included in the production bundle
* They don't need to be since they aren't going to be used in production

## Goal

* Clean up the keg-components exports
* Remove prop-types from the keg-components export

## Updates

* `apps/Sessions.js`
  * Updated container imported by the Session App
* `configs/rollup.config.js`
  * Updated svg import to be `react-native-web`
* `configs/rollup.config.js`
  * Added `docker:build` script in package.json
  * This will build the `keg-sessions.js` export within the docker container
* `scripts/dockerBuild.sh`
  * Script to build the `keg-sessions.js` export within the docker container
* `src/assets/icons/evf/*`
  * Updated all icons to use the `SvgIcon` component from `keg-components`
 * `src/components/button/bookingButton.js`
  * Fixed some circular imports

## Testing

**Test 1**
* Run the docker package => `keg evf pack run docker pull docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-385-remove-deps`
  * Navigate to http://evf-zen-385-remove-deps.local.kegdev.xyz/
  * Ensure the tap loads as expected

**Test 2**
* Pull the PR for the keg-hub updates
  * `keg && keg pr 69`
* Pull this PR for the tap, and run it
  * `keg evf && keg pr 87 && keg evf start --build --sync components:build`
    * We add the `--build` to ensure the babel plugin gets installed
    * We add the `--sync` to create a sync with the keg-components changes from the other PR
       * This ensure we can test the keg-components in the tap, without the `propTypes` in the keg-components build
* Once the container is running, navigate to http://evf-zen-385-remove-deps.local.kegdev.xyz/
  * Ensure the tap runs as expected

**Test 3**
* With the evf container running, in a new tap run the command `keg evf && yarn docker:build`
* This should create a build of the `keg-sessions` component and copy it from the docker container to you host machine
  * **Important** - The terser plugin makes it difficult to see changes to the build
    * For testing, you should comment it out in the taps rollup config, so you can read the output
 * With in the `keg-sessions` build, check that the keg-components propTypes are not included in the build
  * **IMPORTANT** - The components of the `events-force` tap will still have their `propTypes` only the keg-components propTypes will be removed 
* Also check that the following libraries do not exist
  * `axios`
  * `ace`
  * `react-native`
  * `react-native-svg`

**Test 4**
* Run the keg-consumer docker package => `keg consumer pack run docker.pkg.github.com/simpleviewinc/keg-packages/keg-test-consumer:zen-385-remove-deps`
> **IMPORTANT** - The URL does not match the branch name. I built the image then changed the branch name to match the other PRs. 
* Navigate to http://consumer-reducer-sessions-bundle-size.local.kegdev.xyz/
  * Ensure the app runs as expected
  * Ensure no errors show up

**Test 5**
* Attach to the consumer docker container
  * `keg d att package-keg-test-consumer-zen-385-remove-deps`
* Navigate to the keg-sessions node_module
  * `cd /keg/app/node_modules/@keg-hub/tap-evf-sessions`
* Print out the `keg-sessions.esm.js` file
  * Compare that file to the build file from **Test 3**
  * **They should be the same, with the same imports**
  * This confirms the build works with-in the keg-consumer app
> I did the build without `terser` so it would be easy to compare


mpcarolin (changes requested):
Some of the icons don't look right. Mainly they look much larger than normal:
image
image

daniel112:
@mpcarolin  you most likely have to run `keg evf start --build --sync components:install:build`, notice the `install`
the keg-components build failed when i didnt have the `install` which caused the giant icons too

package.json
```diff
@@ -13,6 +13,7 @@
     "type": "git"
   },
   "scripts": {
+    "docker:build": "bash ./scripts/dockerBuild.sh",
```
daniel112, line 16: just to verify
on the build output. it has things for keg-components propTypes but its empty. like this:
image

instead of it showing the properties for the defined PropTypes. Im guessing this is fine. since the actual key properties are removed

daniel112 (approved):
looks good. had 1 comment just for verification

mpcarolin (approved):
Looks good. Daniel was correct -- just needed the install action for the components to get rollup.

PR #89, opened 2020-11-11
Add Reactstrap modal and Remove Styles Parser

lancetipton:
**Ticket**: [ZEN-446](https://jira.simpleviewtools.com/browse/ZEN-446)

## Context

* The modals within the sessions component and the EUs Teams app is not consistent
* We need a way to ensure the modal is the same in use cases
* Moving forward the EU team will pass in a modal component to be used by the sessions component

## Goal

* Switch the modal component. Use a component passed in from the consumer

## Updates

* `apps/Main.js`
  * Added bootstrap styles to the modal looks correct
  * Should only be used in development
* `package.json`
  * Added `bootstrap` and `reactstrap` dev dependencies
  * Fix minor issue with double imports
* `src/components/button/evfButton.js` && `src/components/sessions.js`
  * Removed the use parsed styles code
  * Fixed some imports
  * Added import for the `modalContext`
* `src/components/modals/modalContext.js`
  * Added modalContext file to store the passed in modal component
* `src/containers/rootContainer.js`
  * Added code for passing in the modalComponet, matching to how the EU team will do it
* `src/containers/sessionsContainer.js`
  * Updated to use the `modalContext` provider to set the passed in modal component
* `src/mocks/eventsforce/evfModalBuilder.js`
  * Added demo modal component matching the EU teams component
  * Removed, as it's no longer needed

## Testing

**Test 1**
* Run this docker package url =>`keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:zee-446-react-strap-modal`

* Navigate to this url in the browser => http://evf-zee-446-react-strap-modal.local.kegdev.xyz/
  > The tap is using this updated version of re-theme, with the styles parser removed

* You may see some errors why the docker package is starting.
  * They look like this => http://snpy.in/6roPOD
  * These errors are not related and can be ignored
  * I'm testing updating to Node 14, and there a number of changes that cause these error to show up
  * They do not affect the actual application running

* Ensure
  * no errors are thrown from within the app
  * the app run as expected
* **IMPORTANT** - The modal styles are expected to be incorrect
  * This is due to the updates to the tap's modal using `boot-strap`
  * There is a seperate ticket to fix these issues

**Test 2**
* Go through each modal and ensure it opens as expected
  * Interact with the UI within the modal
  * All functionality should still work as expected


src/mocks/eventsforce/evfModalBuilder.js
```diff
@@ -0,0 +1,37 @@
...
+        isOpen={isOpen}
+        toggle={toggle}
+      >
+        { modalHeader && <ModalHeader>{ modalHeader }</ModalHeader> }
+        { modalBody && <ModalBody>{ modalBody }</ModalBody> }
+        { modalFooter && <ModalFooter>{ modalFooter }</ModalFooter> }
+      </Modal>
+    )
+  }
+}
```
mpcarolin, line 37: Not blocking b/c this is just for development, but:

why not just make this a component?

If it were a component then you could wrap it in `React.memo` and then `rootContainer` wouldn't need the `useMemo` call.

mpcarolin (approved):
Looks good to me. I had one comment, but not blocking.

src/mocks/eventsforce/evfModalBuilder.js
lancetipton: Yah I could do that, because it's just a HOC.
Honestly I just didn't want to overcomplicate it, because the real one is coming from the EU team.
I was trying to make it as simple as possible for them to integrate. 
But you are totally correct the it could be a Component with `React.memo`

src/containers/sessionsContainer.js
```diff
@@ -13,18 +14,25 @@ import { useSelector } from 'react-redux'
  *                                                  - passes back session id and an array of attendee ids
  */
 export const SessionsContainer = props => {
```
daniel112, line 16: minor thing. can you update the jsdoc for the `ModalComponent`

daniel112 (approved):
modal functionality works as expected
1 comment on jsdoc

src/containers/sessionsContainer.js
lancetipton: Updated

PR #90, opened 2020-11-12
Zen 427 group time conflicts

mpcarolin:
**Ticket**: [ZEN-427](https://jira.simpleviewtools.com/browse/ZEN-427)
and [ZEN-458](https://jira.simpleviewtools.com/browse/ZEN-458)

## Context

* currently, the group booking modal does not disable attendee checkbox items if those attendees have a time conflict between the selected session and another session

## Goal

* update the group booking modal to ensure that attendees with time-conflicts are disabled and greyed-out

## Updates

* `src/components/booking/attendeeBookingList.js`
  * updated to disable attendees checkbox items if they have a time conflict
* `src/hooks/booking/useBookingLists.js`
  * updated to ensure that attendees with time-conflicts are never pre-selected by the group booking modal
* `src/hooks/models/attendees/useIsAttendeeDisabledCallback.js`
  * created a more general hook, `useIsAttendeeDisabledCallback`, that checks if the attendee has either a time conflict OR is restricted from booking a session.
* `src/hooks/booking/useBookingTimeConflicts.js`
  * ensured that if `hasTimeConflict` returns a number, that number is a booking conflict id (sometimes this function could short-circuit return a 0 if the booked list was empty, but that could be misinterpreted as a conflict id)
  * also made `hasTimeConflict` work if the bookedId is a number rather than a string
  * updated `getTimeConflicts` to add a conflict id to the conflicts list if that id is 0
* `src/mocks/eventsforce/bookingStates/conflict.js`
  * just created a mock state for testing

## Testing

* `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-427-group-time-conflicts`
* navigate to http://evf-zen-427-group-time-conflicts.local.kegdev.xyz/?state=conflict.group.limited
* select the second session
  * verify that `Mr. Frank Smith` is disabled. This is because Frank Smith is already booked against the first session, which you can see in the test data json. These two sessions have the exact same time block and day.
* now select the first session
  * verify one of the attendees are pre-selected since they are already booked to the session
* select `Mr James` to be booked
* select the `Book Selected` button
* verify that the second session's booking button is now disabled
* if you'd like, try any combination

### DIFF
--- a/src/hooks/booking/useBookingLists.js
+++ b/src/hooks/booking/useBookingLists.js
@@ -1,7 +1,7 @@
 import { useMemo } from 'react'
 import { getExistingWaitIds } from 'SVUtils/booking/getExistingWaitIds'
 import { getExistingBookIds } from 'SVUtils/booking/getExistingBookIds'
-import { useRestrictedAttendeeIds } from './useRestrictedAttendeeIds'
+import { useIsAttendeeDisabledCallback } from 'SVHooks/models/attendees/useIsAttendeeDisabledCallback'
 
 /**
  * @param {boolean} initialCapacityExceedsNeed
@@ -26,21 +26,21 @@ const shouldPreselectAttendees = (
  * @param {Object} session - session object
  * @param {Array<import('SVModels/Attendee').Attendee>} attendees - full list of attendees
  * @param {Array<string>} initialWaitIds - list of ids on the initial waiting list
- * @param {Function} isBookable - cb of form: attendeeId => true/false if bookable/waitListable to the session
+ * @param {Function} isAttendeeDisabled - cb of form: attendeeId => true/false if bookable/waitListable to the session
  * @param {boolean} initialCapacityExceedsNeed - true if the initial capacity exceeds the potential
  * @return {Array<string>} initial booking ids
  */
 const getInitialBookingIds = (
   existingBookIds,
   attendees,
-  isBookable,
+  isAttendeeDisabled,
   shouldPreselect
 ) => {
   // if nobody is on the waiting list, and the capacity of session is greater than the number of attendees,
   // include all the attendees that are bookable
   if (shouldPreselect)
     return attendees.reduce((ids, nextAttendee) => {
-      if (!isBookable(nextAttendee.bookedTicketIdentifier)) return ids
+      if (isAttendeeDisabled(nextAttendee.bookedTicketIdentifier)) return ids
       ids.push(nextAttendee.bookedTicketIdentifier)
       return ids
     }, [])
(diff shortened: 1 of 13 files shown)


src/hooks/booking/useBookingLists.js
```diff
@@ -1,7 +1,7 @@
 import { useMemo } from 'react'
```
daniel112, line 1: marking this file for the unit test that failed
image

src/hooks/models/attendees/useIsAttendeeDisabledCallback.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Helper for determining if an attendee is restricted from booking a session
+ *  - checks both the restricted attendee list and searches for time conflicts with other sessions
+ * @param {import('SVModels/Session').Session} session
+ * @param {Array<import('SVModels/Attendee').Attendee>} attendees
+ * @returns {Function?} a callback of form (attendeeId) -> Boolean, which returns
+ * true if the attendee is disabled (cannot be booked or selected in the group booking modal)
+ */
+export const useIsAttendeeDisabledCallback = (session, attendees) => {
```
daniel112, line 40: for an edge case where every single person in the group is booked against a session. it should disable the booking button on the rest of the sessions that's conflicting similar to how it is for the single booking.

current behavior still enables the booking button, but when you open it, all users are disabled
image

daniel112 (changes requested):
unit test had 1 failing test for me
I also have an edge case that i think we should cover for consistency

src/hooks/booking/useBookingLists.js
mpcarolin: fixed

src/hooks/models/attendees/useIsAttendeeDisabledCallback.js
mpcarolin: Good catch. This had to do with the hook `useGroupCounts`, which was only counting using the restricted attendees. I updated it to factor in the time-conflicted ones as well 👍

lancetipton (approved):
Code looks good. Most items worked as expected, but I did run into something that may be an issue
When clicking on the first session from the test steps 4th bullet point, only one attendee is selected => http://snpy.in/LRYnCT
  * Which I'm pretty sure is the correct behavior
  * But the test says both users should be selected
    * From my understanding, it should only set all users as selected, when no attendees have been booked to the session
    * But because one of the attendees is booked to the session, it does not auto select all of them
  * When I remove the attendee from being booked to that session. Then when I open the modal, all users are auto selected.

So I'm going to approve because I believe everything is working correctly, even though that's not what the test says it should be doing

src/hooks/models/attendees/useIsAttendeeDisabledCallback.js
daniel112: will continue in https://jira.simpleviewtools.com/browse/ZEN-458

daniel112 (approved):

PR #92, opened 2020-11-16
Theme As Functions

lancetipton:
[Semver-Status] patch

> This PR should be reviewed along with [this PR](https://github.com/simpleviewinc/keg-hub/pull/72)

## Context
* Currently there is no way to overwrite the default theme config
* Keg-Components has been updated to allow dynamically setting the Theme
* This allows overwriting the default theme config

## Goal

* Should allow overwriting the default theme config

## Updates

* `apps/Main.js` && `apps/Sessions.js`
  * Moved call to setDefaultTheme to theme file
* Removed `setRNPlatform` call. It's no longer needed
* `src/theme/tapIndex.js`
  * Updated to call theme initialize function

## Testing

* Run the docker package => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:keg-theme-as-functions`
* Navigate to this URL => http://evf-keg-theme-as-functions.local.kegdev.xyz/
  * Ensure everything looks exactly the same as before
  * There should be no difference between the current EVF Tap styles, and the changes from this branch

### DIFF
--- a/apps/Main.js
+++ b/apps/Main.js
@@ -2,7 +2,7 @@ import '../src/theme/theme.config'
 import React, { useState, useEffect } from 'react'
 import { theme } from 'SVTheme'
 import { SafeAreaView, StatusBar } from 'react-native'
-import { getDefaultTheme, setDefaultTheme, ReThemeProvider } from '@keg-hub/re-theme'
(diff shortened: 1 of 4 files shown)


daniel112 (approved):
everything looks normal to me

mpcarolin (approved):
Looks good to me

PR #95, opened 2020-11-17
Zen 419 booking context

mpcarolin:
**Ticket**: [ZEN-419](https://jira.simpleviewtools.com/browse/ZEN-419)

## Context

* the group booking modal needs to manage a lot of state specific to group booking, like the current booking list and the current capacity
* currently, this is stored in the global redux store, and it's managed by many hooks and actions. It's becoming hard to reason about. It also isn't intuitive, because it's stored globally, but it is specific to the group booker and gets reset every time a new group booker opens.
* it would be simpler to manage the group booking state with a react context and a reducer.

## Goal

* Refactor the group booking code to use react context rather than the redux store

## Updates

* `src/actions/session/booking/*`
  * removed several actions we no longer need
* `src/components/booking/attendeeBookingList.js` & `/attendeeCheckBoxItem.js`
  * updated to use the context state
  * simplified the code so that it wasn't calling `useMemo` in a loop
  * removed all initialization code that was happening in the function body through hooks
  * updated footer/header to use the context
* `src/components/booking/groupBookingSection.js`
  * updated jsdocs
* `src/components/modals/groupBooking.js`
  * updated to include the context provider here, to ensure it gets initialized before the body, header, or footer try to use it
  * also to be forward-compatible with Daniel's upcoming changes to the baseModal
* `src/constants/values.js`
  * removed constants we don't need any more
* `src/contexts/booking/groupBookingActionTypes.js`
  * some action type constants specific to the group booking context reducer
* `src/contexts/booking/groupBookingContext.js`
  * exporting the context and a helper hook to access it
* `src/contexts/booking/groupBookingInitialState.js`
  * exports the initial state of the group booking reducer
* `src/contexts/booking/groupBookingProvider.js`
  * the provider, which initializes the context state
  * also defines the reducer actions here
* `src/contexts/booking/groupBookingReducer.js`
  * reducer for handling state mutations
  * currently just handles the state change for updating the booking or waiting list with an attendee id (add or remove)
    * also handles updating the capacity and the "modified" properties
* `src/contexts/booking/useInitialState.js`
  * builds the initial state for the reducer
* `src/hooks/booking/useBookSessionCallback.js`
  * I removed the `useSessionBooking` hook, which did a number of things, and extracted this out of it.
  * this is all we need anymore: just returns the callback for booking a session
* `src/hooks/booking/*`
  * removed hooks we don't need anymore
* `src/hooks/booking/useBookingTimeConflicts.js`
  * linter's work
* `src/mocks/eventsforce/bookingStates/greaterCapacity.js`
  * just another test state for testing a session that is NOT unlimited, but has greater capacity than need.
  * access at: `?state=greaterCapacity.group.limited`
* `tap.js`
  * added an alias to the contexts folder, `SVContexts`

## Testing

`keg evf pack run ﻿﻿docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-419-booking-context`

Then visit [evf-zen-419-booking-context.local.kegdev.xyz](http://evf-zen-419-booking-context.local.kegdev.xyz)

* there are no new features, so we just need to test that the group booking modal works like it did before
* I went through and tested the instructions from this previous PR: https://github.com/simpleviewinc/tap-events-force/pull/63
  * however, the `/test` page doesn't exist anymore, so just use one of the test states, the existing sessions, or update the JSON to work as it describes
* to test the "Greater Capacity than Need" section, visit `?state=greaterCapacity.group.limited`


src/contexts/booking/groupBookingContext.js
```diff
@@ -0,0 +1,5 @@
+import { useContext, createContext } from 'react'
+
+export const GroupBookingContext = createContext()
+
+export const useGroupBookingContext = () => useContext(GroupBookingContext)
```
daniel112, line 5: no issue with this file specifically, I just want to make sure I understand fully.

from just looking at the structure. i was under the impression that `/contexts` contains only context + provider setup.
but it also contains hooks, initialState, actions and reducers.

was the intent just to group related functionality together in one module? 
the pattern just seems a bit weird and disconnected from the rest of our setup.

maybe this is just common context pattern? if so then i'm ok to go with it

daniel112 (commented):
still testing, but just want to post this comment first

src/contexts/booking/groupBookingActionTypes.js
```diff
@@ -0,0 +1,3 @@
+export const GroupBookingActionTypes = {
```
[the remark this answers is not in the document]
mpcarolin: I considered that, but these action types are only used in the `src/contexts/booking/*` files. They wouldn't be used elsewhere, and I didn't want them to be confused with the global redux action types. So I put them here. I'm open to changing that though.

src/contexts/booking/groupBookingContext.js
mpcarolin: From my research, this is definitely a common react pattern. It's a reducer that is specific* to, and provided by, a single context, rather than globally accessible like redux. It seems disconnected because it it's supposed to.

*(yes, technically the context is global in the sense that any component _could_ import it, but that's discouraged, and since it only gets initialized when the group booking modal opens, it would be `undefined` for anyone who tries to access it outside of that window).

src/contexts/booking/useInitialState.js
```diff
@@ -0,0 +1,41 @@
+import { initialState } from './groupBookingInitialState'
```
lancetipton, line 1: Normally I would be all for the name `useInitialState`, but since this is specific to booking, I think it would be important to state that. So I would update the name of this file to `useInitialBookingState`

src/contexts/booking/groupBookingProvider.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {import('SVModels/Session').Session} props.session - the session object currently used in the group booking UI
+ * @param {*} props.children - children
+ */
+export const GroupBookingProvider = ({ session, children }) => {
+  const initialState = useInitialState(session)
+
+  const [ state, dispatch ] = useReducer(groupBookingReducer, initialState)
+  const actions = useActions(dispatch)
+
+  const contextValue = useMemo(() => ({ state, actions }), [ state, actions ])
```
lancetipton, line 50: I'm not sure I understand why this is needed. 
* The state only updates from an action, so it's already memoized in a way, through the `useReducer` hook
* From my understanding, the actions will never change, because the dispatch will should never change.
  * See this [image](http://snpy.in/EljTt1) from the docs
    * Here's the [link](https://reactjs.org/docs/hooks-reference.html#usereducer) if you want to read more

Because the dispatch function never changes, it means there's no need to memorize the actions callback
Furthermore, because the state only changes from action calls, there's no need to memorize it either.

But there could be something I'm missing?

src/contexts/booking/groupBookingActionTypes.js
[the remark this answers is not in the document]
daniel112: there exists some constants thats currently being used for a single use case, so i think it's fine to put your constants in the `/constants` dir

if you want, we can make it more clear on the use case by putting your actionType under a `context` key.
so something like
`ActionTypes.CONTEXT.BOOKING.UPDATE_SESSION_BOOKING` or just `ActionTypes.CONTEXT.UPDATE_SESSION_BOOKING`

src/contexts/booking/groupBookingContext.js
daniel112: I don't have too strong of an opinion on this. but I think at the very least we should separate the different types of functions that you made under `contexts/booking` since not all of them are `context` or `provider`

after discussing further in chat. we want to ensure that the hooks and other utilities inside `contexts/booking` is only ever used within that context, so I'm okay with separating the hooks, actions, reducers inside the `contexts/booking` folder

```
   /context
     /booking
       /hooks
       /actions
       provider.js
       context.js
```

or something like that

src/contexts/booking/groupBookingReducer.js
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * Updates the list identified by listKey using the updateFn
+ * @param {Object} state - state of group booking reducer
+ * @param {string} listKey - key name of the list to update
+ * @param {Function} updateFn - fn of form (currentList) => nextList, returning the next list to reduce to
+ * @return {Object} - next state
+ */
+const updateList = (state, listKey, updateFn) => {
+  const [valid] = validate(
```
lancetipton, line 13: Not sure there's much point in doing this validation here
  * All calls to this method are done internally to this file, which means it's fully managed
  * It only really makes sense to use validation on the exported method in `groupBookingReducer` which calls `updateSessionBooking`
    * You are also doing validation there. And that makes more sense and provides value.
    * The validation here, I think is pointless

lancetipton (changes requested):
Looks really good. I like this so much more compared to before.
I left a few comments and questions. Thanks

src/contexts/booking/groupBookingActionTypes.js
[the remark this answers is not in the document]
lancetipton: Personally, I like the it seperated that way it is. 
The whole group booking section is basically it's own app.
Keeping it separated removes ambiguity, and ensures proper scoping.

That being said, if there are things that are shared it would make sense to join them together, but I don't think group booking actions is one of them.

Just my 2 cents

src/contexts/booking/groupBookingContext.js
lancetipton: Nice, I'm glad ya'll came to this conclusion. Cause that's also what I would have recommended.

src/contexts/booking/groupBookingProvider.js
lancetipton: After discussing this is slack, memoizing the state and actions as a single object makes sense.
This way they can be passed to value prop of the Context together.

src/contexts/booking/groupBookingReducer.js
mpcarolin: It was more intended to help me during development, as I was building this out. But it's served it's purpose so I'll go ahead and remove it.

src/contexts/booking/useInitialState.js
mpcarolin: yep agreed, updated

src/contexts/booking/groupBookingContext.js
mpcarolin: updated with better folder structure

src/contexts/booking/groupBookingReducer.js
```diff
@@ -0,0 +1,131 @@
...
+ * that contains it (either booking or waiting list). Also ensures that
+ * the state's modified and capacity properties are updated accordingly.
+ * @param {Object} state - state of group booker
+ * @param {string} id - attendee id
+ */
+const updateSessionBooking = (state, id) => {
+  const [valid] = validate({ state, id }, { state: isInitialized, id: isStr })
+  if (!valid) return state
+
+  console.log(id)
```
daniel112, line 89: can you remove this?

daniel112 (approved):
looks good. just 1 minor thing not blocking

src/components/booking/groupBooker.js
```diff
@@ -87,14 +51,13 @@ GroupBooker.propTypes = {
...
+const TopSection = ({ styles }) => {
   // use correct wording depending on number of spots remaining
-  const placeText = remainingCount === 1 ? 'place' : 'places'
-
-  const showCount = exists(remainingCount) && remainingCount !== Infinity
-
+  const { state } = useGroupBookingContext()
+  const placeText = state.capacity === 1 ? 'place' : 'places'
+  const showCount =
+    state.showCapacity && exists(state.capacity) && state.capacity !== Infinity
```
lancetipton, line 60: Pretty sure the first `state.showCapacity` is not needed here. Is there a reason for it?

lancetipton (changes requested):
Looks good. I really like the folder structure, and everything is very clean. Great work
* Just one minor thing to look at. Thanks

src/components/booking/groupBooker.js
mpcarolin: I actually just added it to fix a bug I introduced with this refactor.

The spec actually requires the capacity count to be hidden if:
- the capacity is unlimited 

in which the session capacity isn't "unlimited", but it's still greater than the number of "bookable" attendees

src/contexts/booking/groupBookingReducer.js
mpcarolin: thanks. updated

lancetipton (approved):
Great work. Looks good. Thanks

PR #96, opened 2020-11-18
Zop 505 fix modal style

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZOP-505)
> Note: i rebased with michaels branch https://github.com/simpleviewinc/tap-events-force/pull/95 because of the group booking modal changes to avoid major conflicts

## Context

* We need to fix the styles for our modal now that we're using reactstrap modal

## Goal

* update the styling for the BaseModal and all other modals we're using so it looks like how it did before in terms of the **Body** and **Footer** content

## Updates

   * updated to take in `Body` and `Footer` prop as component
   * removed `Header` component and just use the `title` prop string as the header
   * removed any base styling, since evf can now control the styling fully

    * exported/renamed `GroupBooker` to `GroupBookerBody`
    * exported the `GroupBookerFooter`

* `src/components/modals/alert.js`
* `src/components/modals/groupBooking.js`
* `src/components/modals/presenterDetails.js`
* `src/components/modals/sessionDetailsModal.js`
   * updated to pass in `Body` and `Footer` component accordingly

* `src/mocks/eventsforce/evfModalBuilder.js`
    * added mock styling to the `headerModal` just so it looks good during development

* `src/theme/components/button/evfButton.js`
   * check icon was tiny. fixed it

* `src/theme/components/modals/*`
   * removed uneeded styles

* run `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zop-505-fix-modal-style`
* navigate to http://evf-zop-505-fix-modal-style.local.kegdev.xyz/
* Modals to test:
    * Filter
    * Presenter details
    * session details
    * group booking
    * alert (modify the `alert` prop through JSON editor)
* Verify that the `Body` and `Footer` styling looks better than how it currently is at https://simpleviewinc.github.io/keg-test-consumer/
     * no extra padding on the body and footer
     * Header text should be black and unstyled (we no longer manage the styling on the header)
     * there should no longer exist an `X` icon on the header (evf will manage this too)

### Check Icon
* Change `Booking State` to `select-single-check` > Apply Changes
* Verify that the check icon is not super tiny
    * should look like this [image](https://user-images.githubusercontent.com/3317835/99595298-b3f41480-29b1-11eb-99c7-2e229f191f05.png)


src/theme/components/modals/groupBookingModal.js
```diff
@@ -63,43 +54,36 @@ export const groupBookingModal = {
             },
           },
         },
-        bottomSection: {
+      },
```
lancetipton, line 57: Not sure where else this can go, but we should move the on `waiting list` text box to the right a bit. That way it's not right up on the Users name => http://snpy.in/trifLP
Needs a little space

lancetipton (commented):
Looks good. Just one minor style request

src/theme/components/modals/groupBookingModal.js
daniel112: sure i can add some padding

daniel112: added 10 paddingRight to the text. so it wraps sooner if its close
image

lancetipton: Fix looks great. Thank you sir

lancetipton (approved):
Looks good. Thanks

mpcarolin (approved):
Looks good

PR #97, opened 2020-11-18
Type definition review - Optional

lancetipton:
**Ticket**: [ZEN-449](https://jira.simpleviewtools.com/browse/ZEN-449)

## Context

* Events Force has sent us a types definition file which includes some optional props
* We need to ensure all optional props are not required within the Sessions Component

## Goal

* Ensure optional props do not throw errors when not passed

## Updates

* `src/actions/attendees/initSortedAttendees.js`
  * Made `tickets` && `bookedTickets` optional
* `src/actions/attendees/setAttendeesByTicket.js`
  * Added validation helper, that only validates when the passed in prop exists
  * Otherwise sets it to an empty object
  * Added defaults for the alert prop
* `src/mocks/eventsforce/bookingStates/index.js` && `src/mocks/eventsforce/bookingStates/optionalTickets.js`
  * Added extra booking state to test optional props

## Testing
* Run the docker package => `keg evf pack run docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-449-type-def-review`
* Navigate to http://evf-zen-449-type-def-review.local.kegdev.xyz/
* Open the testData editor, and switch to `optionalTickets-group-optionalTickets`
  * Looks like this => http://snpy.in/APSYaq
* Hit the Apply Changes button
* Open the testData editor again and inspect the test data, and ensure the `tickets` && `bookedTickets` keys do not exist
  * Looks like this => http://snpy.in/dz2cn6
* Check the browser inspector, and ensure no errors were logged
* Ensure the site works as expected


src/actions/attendees/setAttendeesByTicket.js
```diff
@@ -1,26 +1,38 @@
 import { Values, ActionTypes } from 'SVConstants'
 import { dispatch } from 'SVStore'
-import { isArrMap, validate } from '@keg-hub/jsutils'
+import { isArrMap, validate, noOpObj } from '@keg-hub/jsutils'
```
daniel112, line 3: marking a random file for comment.  i wish github would make it so you can  have a reply chain on PR comment and not just file comments..

Even though these 2 props are optional. not passing them in makes Group Booking Modal not display any data.
I just want to verify that this is expected. so these props are optional, but also technically required for group booking modal to work? @mpcarolin 
```Javascript
  bookedTickets?: EfBookedTicketInterface[] // Used to link the bookedTicketIdentifier for the attendee to the actual ticket
  tickets?: EfTicketInterface[] //Used to get the name of the tickets
​
```

daniel112 (commented):
tested for these optional props:
```Javascript
  bookedTickets?: EfBookedTicketInterface[] // Used to link the bookedTicketIdentifier for the attendee to the actual ticket
  tickets?: EfTicketInterface[] //Used to get the name of the tickets
​
  onSessionBookingRequest?: (
    sessionIdentifer: EfSessionIdentifier,
    attendeeBookedTicketIdentifiers: EfBookedTicketIdentifier[]
  ) => void
​
  alert?: EfAlert
```

App still works and does not throw error or break

got 1 question about group booking

src/actions/attendees/setAttendeesByTicket.js
mpcarolin: Yes, without `bookedTickets` and `tickets`, the group booker has no way to determine which groups the attendees should be organized into.

This optionality only makes sense when the app is in single-booking mode. In that case, the group booking modal wouldn't open, so they wouldn't be needed.

daniel112 (approved):

mpcarolin (changes requested):
Passed the testing instructions.

However, if you navigate directly to: http://evf-zen-449-type-def-review.local.kegdev.xyz/?state=optionalTickets-group-optionalTickets, then attempt to open a group booking modal, the app will throw and error and crash (due to undefined tickets). 

This doesn't happen in the test-case where you switch to the state using the apply button, because the `bookedTickets` and `tickets` categories actually aren't getting removed from the store for some reason. You can see in the redux dev tool that both categories are still there. The group booker is empty only because the `attendeesByTicket` map does get properly set to an empty object.

By navigating directly to that state, the `bookedTickets` and `tickets` categories never get set, though, and a few functions still seem to depend on them.

So, there seems to be a separate issue in our `mapSessionsInterface` function that isn't actually overwriting categories if they change to undefined.

For this ticket, I think that we should just prevent it from crashing when this data state arises, even if it's not a state that they should be sending us.

mpcarolin (approved):
Looks good

PR #103, opened 2020-12-14
ZEN-469: undefined categories

mpcarolin:
**Ticket**: [ZEN-469](https://jira.simpleviewtools.com/browse/ZEN-469)

## Context

* There is a bug in the tap: whenever a prop change occurred that omitted a previously-included category, that change would be ignored by the tap.
  * example: changing the props from: 
```js
{
   attendees: [ ... ],
   sessions: [ ... ]
}
```
to 
```js
{
   attendees: [ ... ],
}
```
would result in the old sessions remaining in the items store.

## Goal

* Fix the bug

## Updates

  * the bug mainly originates from here. The code was looping over the `props` to dispatch updates. But if a category key was removed entirely, then that loop would not know to update that category. I updated it to instead loop over the expected categories, looking for them in the props.
* `src/actions/session/setAgendaSessions.js`
  * just updated to use `validate` for useful error logging
* `src/components/dates/updateDayButton.js` and `src/components/sessions.js` and `src/hooks/models/useAgenda.js`
  * fixed another bug that let the user increment the current day infinitely if the agenda days list was empty
* `src/constants/values.js`
  * split up categories into two objects: internal and evf (props) categories. This way, we can loop over just the evf categories that we expect in `mapSessionsInterface`. The `categories` key still exists and just merges these two object together.
* `src/mocks/eventsforce/bookingStates/*`
  * ensured all the categories had default values present
  * added booking testing states

## Testing

* `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-469-undefined-categories`
* Navigate to http://evf-zen-469-undefined-categories.local.kegdev.xyz/
* Switch to the testing state `noData - noSessions` using the dropdown 
  * or just delete the `sessions` key and value from the test data editor (make sure you delete the entire entry)
* Hit `Apply Changes`
* Verify that the sessions disappear, but that the app continues to behave normally
  * console errors are expected -- we deleted an expected prop, so warnings should still be logged
* Test deleting any category you want, verifying that the items state updates accordingly


daniel112 (approved):
testing looks good.
there were a few state that retained some of the values after clearing all the data (modals, attendeesByTicket, restrictedAttendeeIds)
but after discussing it with @mpcarolin . this is a non issue as those states can't be utilized if certain props are empty

lancetipton (approved):
Looks good, and works as expected. 
Thanks

PR #104, opened 2020-12-15
ZEN-464: Removed label css

mpcarolin:
**Ticket**: [ZEN-464](https://jira.simpleviewtools.com/browse/ZEN-464)

## Context

* Early on in this project, we made a script that generates an events force theme using a css file as input
* This was done under the assumption that evf would send us their CSS file(s)
* This is not the case, and we have since implemented different ways of handling css in the keg, so this generated css is no longer needed

## Goal

* Remove the generated theme file and scripts that created it
* Updated the label components to stop depending on that theme
  * all the labels should only have a gray background color, since that color will be defined by evf

## Updates

* `package.json`
  * removed command and dependency only needed for the generated theme
* `scripts/*`
  * removed scripts for generating the theme
* `src/components/labels/labelButton.js` and `src/components/labels/labelTag.js`
  * removed references to the `eventsForce` theme
* `src/theme/*`
  * removed generated file and its dependents' import statements

## Testing

* `keg evf pack run package=docker.pkg.github.com/simpleviewinc/keg-packages/tap:zen-464-remove-labels-css`
* verify for both desktop and mobile web:
  * the tap runs normally
  * all labels / label-tags are gray in color
  * labels still have the right evf class (it should look like this: `ef-agenda-label-<id>`)'
  * in the filter modal, label buttons still have our internal styles (i.e. backgrounds should dim on hover)


src/components/labels/labelTag.js
```diff
@@ -14,11 +14,7 @@ export const LabelTag = ({ label = {}, styles }) => {
   return (
     <View
       className={[ 'ef-label-tag', label.className ]}
-      style={[
-        theme.get(`eventsForce.labels.${label.className}`),
-        theme.get('labelTag.main'),
-        styles,
-      ]}
+      style={[ theme.get('labelTag.main'), styles ]}
```
daniel112, line 17: marking file for ref
on mobile, should we default labelTag color to gray too?

image

daniel112 (commented):
looks good. i posted 1 question

daniel112 (approved):

src/components/labels/labelTag.js
```diff
@@ -1,24 +1,44 @@
...
 import { useTheme } from '@keg-hub/re-theme'
+import { noPropObj } from '@keg-hub/jsutils'
+
+/**
+ * @param {Object} styles - custom styles
+ * @return {Object} styles for the LabelTag component
+ */
+const useLabelTagStyles = styles => {
+  const theme = useTheme()
+  const mainStyles = theme.get('labelTag.main')
```
lancetipton, line 13: The theme hardly ever changes.
Which means, we could move these `theme.get` calls into the `useMemo` hook below.

Then add the theme as a dependency.
This way, `theme.get` calls only happen when the theme or styles change

It's a small performance boost, but every little bit helps.

lancetipton (approved):
Looks good. One minor comment, but it's not blocking. Just a suggestion. 
Thanks

PR #105, opened 2020-12-16
Version Display

mpcarolin:
**Ticket**: [ZEN-509](https://jira.simpleviewtools.com/browse/zen-509)

> [This PR](https://github.com/simpleviewinc/keg-hub/pull/103) for the hub will be merged first

## Context / Goal

* to help inform devs (and anyone testing the consumer) of the current version of the tap, we should show the version of the tap in the consumer that also links to the release notes
* I thought this should go in the tap, so that the events-force team could enable it on their end too, for their testers.

## Updates
* `configs/rollup.config.js`
  * updated to define replacement definitions from the keg envs
* `package.json`
  * added `print:envs` yarn script
* `src/containers/sessionsContainer.js`
  * updated to use `VersionDisplay` component
* `tap.js`
  * updated with `envs` object

## Testing
Follow testing for this PR: https://github.com/simpleviewinc/keg-hub/pull/103


src/components/meta/versionDisplay.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import { Button } from '@keg-hub/keg-components'
+import { colors } from 'SVTheme/colors'
+import tapPackage from '../../../package.json'
```
lancetipton, line 4: This is going to include the entire `package.json`. Which can be bad for a number of reasons. 

* A better way would be to do it through ENV's
* We could add a field to the `tap.js` which defines ENVs to be added to the build process, and exposed to the frontend 
  * You could just do it manually in the keg-core webpack config
      * Using either the `DefinePlugin` || `EnvironmentPlugin`
      * This might be a better option because it adds more customization   
      * This would also allow you to create some shortcuts for accessing the `package.json` values
  * Here's a couple example I found online that could be helpful
    * https://alxmrtnz.com/thoughts/2019/03/12/environment-variables-and-workflow-in-expo.html
    * https://docs.expo.io/guides/environment-variables/

  * They would also need to be added to the rollup config using the `plugin-replace`
    * Exactly the same as how we do handle the other envs
    * Only pull in the `tap.js`, and get the ENV list
      * Check if any have a reference to the `package.json`
        * Something like
        ```js
		  {
			// ...other envs to pass on
		    TAP_VERSION: 'package.version',
		    TAP_URL: 'package.homepage'
		  }
		```
      * `package` could be a shortcut to mean the taps `package.json` file
      * Then loop over them and get the values
      * Then add the values to the plugin-replace initialization
      ```js
        // At top of file import the package.json
        import tapPackage from '../../../package.json'

        // In the plugin-replace initialization
        {
          // ...other replace envs
          'process.env.TAP_VERSION': '<tap-version-here>',
          'process.env.TAP_URL': '<tap-url-here>',
        }
      ```
	
* Just make sure the are the same for webpack and rollup
* This would make it much more dynamic, and keep the bundle size down by not including the entire `package.json`

lancetipton (changes requested):
Looks great and works well.
I left a pretty long comment about it. Let me know your thoughts.

lancetipton (approved):
Code looks good, and everything worked. 
Thanks

daniel112 (approved):
all testing passed

PR #120, opened 2021-02-09
ZEN-541 Day Name Text

daniel112:
**Ticket**: [issue_id](https://jira.simpleviewtools.com/browse/ZEN-541)

## Context

* EVF changed the behavior of the dayToggle again. there must be no calculated rendering of the day date. It must come directly from the dayName json value for each day.

## Goal

- add prop dayName to agendaDay
- use agendaDay.dayName as the label header instead of calculating it

## Updates

* `src/components/sessionsList/sessionsList.js`
   * update to not calculate dayName but instead use the JSON value
* updated the `AgendaDay` model and style slightly

## Testing

* run `keg evf pack run evf:zen-541-dayname-label`
* navigate to http://evf-zen-541-dayname-label.local.kegdev.xyz/
* Open the JSON editor
* Update the `dayName` properties
    * [image](https://user-images.githubusercontent.com/3317835/107407890-79d57300-6ac7-11eb-8895-5274723bd82a.png)
* hit `Apply Changes`
* verify that the dayToggle text changes based on the current day.dayName
* Verify it truncates when the text is long enough

### DIFF
--- a/src/components/sessionsList/sessionsList.js
+++ b/src/components/sessionsList/sessionsList.js
@@ -1,7 +1,7 @@
 import { format, parseISO } from 'date-fns'
 import { isMobileSize } from 'SVUtils/theme'
 import { useTheme } from '@keg-hub/re-theme'
-import { reduceObj, noPropArr } from '@keg-hub/jsutils'
+import { reduceObj } from '@keg-hub/jsutils'
 import React, { useMemo, useCallback } from 'react'
 import { SessionsDivider } from './sessionsDivider'
 import { useAgenda } from 'SVHooks/models/useAgenda'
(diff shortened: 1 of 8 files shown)


src/components/sessionsList/sessionsList.js
```diff
@@ -115,6 +115,21 @@ const useOnScrollChange = (sections, currentDay, onDayChange) => {
...
 }
 
+/**
+ * Gets the text to display as the day name
+ * @param {Array} agendaDays
+ * @param {Number} currentDay
+ *
+ * @returns {string}
+ */
+const getDayName = (agendaDays = noPropArr, currentDay) => {
```
lancetipton, line 125: seems like this should be a hook, something like `useDayName`, then wrap it in a `useMemo`

lancetipton (changes requested):
Looks good. One thing, looks like a `useMemo` hook should be used.

src/components/sessionsList/sessionsList.js
daniel112: I moved this to `useAgenda` since it's related and gets the store items I need

mpcarolin (approved):
Looks good to me

lancetipton (approved):
Looks good, thanks

PR #123, opened 2021-03-12
ZEN-486: Fix for Labels Cut Off

mpcarolin:
**Ticket**: [ZEN-486](https://jira.simpleviewtools.com/browse/ZEN-486)

## Context/Goal

* Currently, in the session details modal, when there are too many labels to fit the width of the modal, they get cut off
* Instead, we they should wrap

## Updates
* `src/components/modals/sessionDetailsModal.js`
  * `SessionDetailsModal` was duplicating the logic handled by `LabelList`, which already does flex-wrap
  * so this update simply updates the component to use `LabelList`
* `src/mocks/eventsforce/testData.js`
  * added more labels so that we can test a large number of labels (the ticket instructs you to try 15 labels, so that's how many I updated the data to have)
* `src/theme/components/labelList.js`
  * did some minor updates to the styles of `labelList`
  * moved `flexWrap` to work in mobile styles as well
  * removed an unnecessary `maxWidth: 14` style that made the label buttons look bad at small sizes
    * this wasn't strictly necessary, because the label buttons usually don't appear in mobile sizes anyway (instead, the label tags do)
    * but it handles one edge case: when you've already opened the session details modal and THEN you shrink the screen 

## Testing

* `keg evf pack run package=ghcr.io/simpleviewinc/evf:zen-486-labels-css-fix`
* Open the session details modal for the first session on the first day
  * This one is already loaded with 15 labels
* Verify that the labels wrap correctly
* Now, shrink the width of your window further, and ensure they still wrap properly
* Also, verify that the other instances of label lists look normal:
  * filter modal
  * session grid session items
    * on desktop
    * on mobile


lancetipton (approved):
Looks good. Everything worked as expected.

daniel112 (approved):

PR #124, opened 2021-03-19
EVF Button Changes

lancetipton:
**Ticket**: [ZEN-556](https://jira.simpleviewtools.com/browse/ZEN-556)

## Context
* See the [Tech Spec here](https://docs.google.com/document/d/1oTOhGc1fpG0VhqXTq4ZumceZWoi1ln17wFVxG9SmlDE/edit)
* Events Force is stating that they need more customization over the Button within the sessions component
* Based on their explanation, they are having issues with styling it properly
* They would like to follow a process similar to the updates we did with the modal component, as seen in these tickets
  * https://jira.simpleviewtools.com/browse/ZEN-446
  * https://jira.simpleviewtools.com/browse/ZEN-502
* Basically, they would pass us a Button component, and we would add some logic to it, and pass it back to them

## Goal

* Allow the evf-button to work like the Modal component and be passed in externally
* Update all references to the button to use this custom evf-button instead of the current button
* Like the modal component, use the bootstrap button when in a development environment
* It should not load the bootstrap button in the production build
* Clean up the button styles

## Updates
* Added Button Component prop
  * Allows passing in a custom button component
  * Updated the modal context to be a Components context
    * Now stores both modal and button components
  * Rewrote button styles
* Added buttonType prop to evf button matching the Tech-Spec requirements
  * Allowed types are `selectSession`, `modalPrimary`, `modalSecondary`

## Testing

* Run the docker packages
  * EVF -  `keg evf pack run evf:zen-556-evf-button-changes`
  * Consumer - `keg consumer pack run consumer:zen-556-evf-button-changes`

> **Notes**
> Please test on both images.
> There were a lot of changes done to keg-hub and this is the first build since those changes were merged

* **Important**
  * There may be some cases where the button does not look correct
  * This is because the root button does not have any styles
  * The EU team is handling that
    * One example of this is the button height in the modal compared to the height in the session list
    * The EU team now controls this
  * We are only responsible for the children

**Test 1**
* Validate that the button children still work as expected
  * The buttons children should still look and act the same
  * The Root Button component should be using the reactstrap button component
* Updates were made to the modal provider
  * So ensure the modal component still displays and closes as expected

**Test 2**
  * Ensure the button icons still render as expected
    * Check for proper alignment
    * Check that they still display for each edge-case defined in the booking state
    * Check both check-mark and the Digit icon variations

**Test 3**
* Ensure all button actions still work as expected
  * Ensure the pending spinner icon still renders as expected
    * When you click book selected, the spinner should display and rotate
  * Modal cancel / book selected
  * Update the mock data to be some variation of single attendee booking state
    * Ensure the check-mark icon still displays when selecting a session


src/mocks/eventsforce/evfButton.js
```diff
@@ -0,0 +1,50 @@
...
+  children,
+  disabled,
+  buttonType,
+  className,
+  onClick,
+  ...props
+}) => {
+  return (
+    <Button
+      className={className}
```
daniel112, line 41: idk if this is expected now or not, but it seems like the className is passed in here, but I don't actually see it on the DOM?
example: on the group booking modal - https://github.com/simpleviewinc/tap-events-force/blob/793406cf8051c573dc73f519062d8c422b0bc444/src/components/booking/groupBooker.js#L112

it has the className `ef-cancel-session-button`, but I don't see that on the DOM
- [image](https://user-images.githubusercontent.com/3317835/111847639-730bfe00-88c6-11eb-8b85-df906dc5b0fa.png)

src/mocks/eventsforce/evfButton.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * This is a demo button component. The real component gets passed in from the Sessions Component consumer
+ * This is basically just a HOC, but called as a function directly
+ *
+ * It will return a functional component to be passed into the Session Component as a prop
+ *
+ * This allows the EU team to define the button in what ever format is needed, including passing in the props
+ * While at the same time, it allows the session component to define the content of the button
+ */
+export const EvfButton = ({
```
daniel112, line 31: All the button child states looks good. In mobile view, I see this odd behavior where the click event passes through the button as well.

- clicking the EvfButton on mobile view will collapse the drawer && open up the modal
new_button

src/contexts/components/componentsProvider.js
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * ComponentsProvider for the components passed as props context.
+ * @param {Object} props
+ * @param {Object} props.components - Custom component to use within the app
+ * @param {Object} props.components.Modal - Custom modal component
+ * @param {Object} props.components.Button - Custom button component
+ * @param {*} props.children - children of the context
+ */
+export const ComponentsProvider = ({ ButtonComponent, ModalComponent, children }) => {
```
mpcarolin, line 12: I like this. This is what I was thinking we should do as well.

src/contexts/components/componentsProvider.js
```diff
@@ -0,0 +1,23 @@
+import React, { useMemo } from 'react'
+import { ComponentsContext } from './componentsContext'
+
+/**
+ * ComponentsProvider for the components passed as props context.
+ * @param {Object} props
+ * @param {Object} props.components - Custom component to use within the app
+ * @param {Object} props.components.Modal - Custom modal component
+ * @param {Object} props.components.Button - Custom button component
```
mpcarolin, line 9: props should be updated to ModalComponent and ButtonComponent

mpcarolin (approved):
Looks good

src/mocks/eventsforce/evfButton.js
lancetipton: According to the docs here => https://reactstrap.github.io/components/buttons/
It should accept the className prop. Seems like a bug in the bootstrap code.
Doesn't really matter though, because the UK team will be owning that, so I'm not really worried about it

src/mocks/eventsforce/evfButton.js
lancetipton: Yah, I just fixed this. I'll push a new package in just a sec

daniel112: looks good

daniel112 (approved):

lancetipton:
Holding off to merge until https://github.com/simpleviewinc/tap-events-force/pull/125 is ready

PR #128, opened 2021-04-20
ZEN-543: Develop Tag

mpcarolin:
**Ticket**: [ZEN-543](https://jira.simpleviewtools.com/browse/ZEN-543)

## Context

* As part of ZEN-543, we determined we should change the default docker image tag from "master" to "develop" for clarity and consistency with the git branches

## Goal

* Update all references of the old "master" to "develop"
* Ensure building images uses "develop"

## Updates
* `container/Dockerfile`
  * updated keg base image to be `tap:develop`
* `container/values.yml`
  * updated tag to "develop"

## Testing

* follow the "setup" section in the testing notes here: https://github.com/simpleviewinc/keg-cli/pull/59
  * you don't have to delete all images, though, just `evf`
* `keg evf build`
* verify it is tagged with "develop"

### DIFF
--- a/container/values.yml
+++ b/container/values.yml
@@ -26,7 +26,7 @@ env:
   KEG_CONTEXT_PATH: "{{ cli.taps.evf.path }}"
   
   # Image to use when building evf
-  KEG_BASE_IMAGE: ghcr.io/simpleviewinc/tap:master
+  KEG_BASE_IMAGE: ghcr.io/simpleviewinc/tap:develop
 
   # Image to use when running evf
   KEG_IMAGE_FROM: ghcr.io/simpleviewinc/evf:develop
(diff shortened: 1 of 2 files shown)


container/values.yml
```diff
@@ -29,7 +29,7 @@ env:
   KEG_BASE_IMAGE: ghcr.io/simpleviewinc/tap:master
```
lancetipton, line 29: Why not update this one?

lancetipton (changes requested):
Something seems off. For the keg-herkin PR you updated the `KEG_BASE_IMAGE` env to point to develop.
But on this one you left it to master. Any reason for that?

container/values.yml
mpcarolin: just an oversight, thanks for catching it.

lancetipton (approved):
Looks good

PR #131, opened 2021-04-30
ZEN-543: Rebuild Workflow

mpcarolin:
**Ticket**: [ZEN-543](https://jira.simpleviewtools.com/browse/ZEN-543)

## Context/Goal

* To save developers from a lot of manual, error-prone work, we've decided to make a github action for building and pushing taps repos' docker images when a push to develop is made 

## Updates
* `.github/workflows/imageDeploy.yml`
  * created the workflow that leverages the tap action
* `tap.js`
  * added an alias key to the `tap.js` file, so that you can run `keg tap link` in the evf directory and it will link that alias automatically

* additionally, you should review the code at the action repo itself: https://github.com/simpleviewinc/tap-build-action

## Testing
* verify that this action succeeded: https://github.com/simpleviewinc/tap-events-force/actions/runs/798249626


tap.js
```diff
@@ -14,6 +13,7 @@ const pathToEntryPoint = isWeb
 
 module.exports = {
   name: 'events-force-x5',
+  alias: 'evf',
```
lancetipton, line 16: Can you move this inside of the `keg` property object.

.github/workflows/imageDeploy.yml
```diff
@@ -0,0 +1,17 @@
...
+  test-build-push:
+    runs-on: ubuntu-latest
+    name: Build tap image
+    steps:
+    - name: Tap Build and Push
+      id: tap-build-push
+      uses: simpleviewinc/tap-build-action@0.0.3
+      with:
+        cli_git_branch: ZEN-543-tap-workflow
+        token: ${{ secrets.KEG_ADMIN_TOKEN }}
```
lancetipton, line 17: I thought we checked and validated that the `secrets.GITHUB_TOKEN` env works here?

lancetipton (changes requested):
Looks good, two small things. Thanks

tap.js
mpcarolin: updated

.github/workflows/imageDeploy.yml
mpcarolin: That was for `keg-hub`'s workflow.

When I attempted it here, it did not work: 
https://github.com/simpleviewinc/tap-events-force/actions/runs/797803949
* in this example, if you examine the workflow file, the token parameter is omitted, but the action defaults to using `${{github.token}}`

lancetipton (approved):
Looks good

PR #135, opened 2021-08-19
ZEN-649: Checkbox Accessibility Updates

mpcarolin:
**Ticket**: [ZEN-649](https://jira.simpleviewtools.com/browse/ZEN-649)

## Goal

* Use a label with the attendee checkbox item and link it to the input element, to improve accessibility

## Updates
  * updated the `id` to be more clear
  * extracted the label component to a separate file
* `src/components/booking/attendeeCheckboxLabel.js`
  * this is the extracted label component
  * updated it to use a `label` instead of `Text` (on web)
  * it now handles both the waiting and non-waiting state 
  * refactored it to use styled components (where possible -- to do more, we need to update `reStyle` to support size keys)
  * this is just removing the modified asterisk that they requested we remove. It's technically handled by another ticket but I knew exactly what to do so I just did it
* `src/components/checkbox/evfCheckbox.js`
  * updated to pass down the id prop
* `src/components/form/label/label.{platform}.js`
  * This is just a simple label element
  * on native, it's just `Text`
  * on web, it's a `label`
    * I did this because while the `react-native-web` `Text` element can support being a `label` element for accessibility, it does not accept the `for`/`htmlFor` attribute
* `src/theme/components/booking/attendeeCheckboxItem.js`
  * removing styles that are now defined in styled components

### Setup
* Open this branch in a codespace. Read the guide [here](https://docs.github.com/en/codespaces/developing-in-codespaces/using-codespaces-for-pull-requests#opening-a-pull-request-in-codespaces)
  * I could send a link to my public codespace, but it stops itself after inactivity so it might be down if I did
* `yarn install`
* `yarn web`

### Instructions
* For both mobile and desktop web:
  * Select the booking button for the session titled `Session on day 2 - limited capacity`
  * Verify that the checkboxes and their associated attendee names are all styled correctly (see the test consumer: https://simpleviewinc.github.io/keg-test-consumer/ for comparison)
  * Inspect the DOM, and verify that a `label` element is used to display the attendee name
  * Verify that it has a `for` attribute, and take note of its value
  * Verify that the value matches the associated checkbox's `id` attribute
  * Also verify that you see no red asterisk next to he `Select sessions for your group` text when you first open the modal

### DIFF
--- a/src/components/form/label/label.native.js
+++ b/src/components/form/label/label.native.js
@@ -1,3 +1,4 @@
-import { Text } from '@keg-hub/keg-components'
+import { Label as KegLabel } from '@keg-hub/keg-components'
 
-export const Label = Text
+export const Label = KegLabel
+
--- a/src/components/form/label/label.native.js
+++ b/src/components/form/label/label.native.js
@@ -1,0 +1,3 @@
+import { Text } from '@keg-hub/keg-components'
+
+export const Label = Text
--- a/src/components/form/label/label.web.js
+++ b/src/components/form/label/label.web.js
@@ -1,6 +1,24 @@
-import { unstable_createElement as createElement } from 'react-native-web'
+import React from 'react'
+import { Text } from '@keg-hub/keg-components'
+import { reStyle } from '@keg-hub/re-theme/reStyle'
+import PropTypes from 'prop-types'
 
+const StyledLabel = reStyle('label')(() => ({ margin: 0, cursor: 'pointer' }))
+
+/**
+ * Label component
+ * @param {*} props
+ * @returns
+ */
 export const Label = props => {
-  const { for: htmlFor, ...rest } = props
-  return createElement('label', { htmlFor, ...rest })
+  const { children, onPress, ...rest } = props
+  return (
+    <Text onPress={onPress}>
+      <StyledLabel {...rest}> { children } </StyledLabel>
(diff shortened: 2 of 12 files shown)


src/components/form/label/label.native.js
```diff
@@ -0,0 +1,3 @@
+import { Text } from '@keg-hub/keg-components'
```
lancetipton, line 1: What not use the exported label when on native?

src/components/form/label/label.web.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import { Text } from '@keg-hub/keg-components'
+import { reStyle } from '@keg-hub/re-theme/reStyle'
+import PropTypes from 'prop-types'
+
+const StyledLabel = reStyle('label')(() => ({ margin: 0, cursor: 'pointer' }))
```
lancetipton, line 6: Ahh, very nice. I really like this solution for allowing the extra props to be passed through.
It's very minimal, and contained

lancetipton (approved):
Looks super clean well done. 
I only had one comment, but it's non-blocking. Great work

src/components/form/label/label.native.js
mpcarolin: Agreed. Updated

PR #144, opened 2021-09-30
More test folders

mpcarolin:
## Context
Fixing issue where keg-herkin couldn't start due to missing test folders


lancetipton (approved):
sweet

PR #152, opened 2021-10-21
Zen 629 price in select

subaya710:
**Ticket**: https://jira.simpleviewtools.com/browse/ZEN-629

## Context

To change the behavior of the price functionality 

## Goal

* Ensure the app matches templates as provided in https://www.figma.com/file/aLesACdNG3CvInUVuFm4hO/Session-selection---final
* Ensure price is available in the select button

## Updates

* Updated BookingButton.js to ensure text and price is displayed according to the rules :
 a. If the session has a price and is a SELECT state, the text displayed should change to BUY.
 b. If the session does not have a price and is a SELECT state, the text displayed should remain SELECT.
 c. If there is no cost , the button should not display any price or alt text like FREE
 d. If a session has a price but is ANY other state (e.g WAITING_LIST) no price should be displayed

## Testing

* Setup
  * Run the command => keg evf pack run evf:zen-629-priceinselect
  * Navigate to http://evf-ZEN-629-PriceInSelect.local.keghub.io
  Note : The current default test data should be enough to validate this ticket

* Testing all requirements
  * Ensure that for a select session with a price , the text on the button appears as 'BUY <price>'. 
  e.g : BUY $923
  * Ensure that for a select session without a price , the text on the button appears as just 'SELECT'
  * Ensure no FREE label appears anywhere for any state
  * Ensure for a waiting list , even though it has a price , no price is displayed . i.e. text should only be 'WAITING LIST' regardless of whether the session has a price or not


mpcarolin (changes requested):
Just setting this to request-changes for the planned integration test

subaya710:
> Just setting this to request-changes for the planned integration test

Hey Michael . Integration tests have been added . Please reference just the src/components/__tests__/bookingButton.integration.js file for your review since that was the only one updated since your last review. Thanks

src/components/__tests__/bookingButton.integration.js
```diff
@@ -0,0 +1,41 @@
...
+const getBuyBtnWithTextBuyAndPrice = () =>
+  screen.getByRole('button', { name: 'BUY $923' })
+
+const getSelectBtnWithOnlyText = () =>
+  screen.getByRole('button', { name: 'SELECT' })
+
+const getWaitingListBtnWithOnlyText = () =>
+  screen.getByRole('button', { name: 'WAITING LIST' })
+
+describe('Booking button with price integration test', () => {
```
mpcarolin, line 18: I would just name this test "Booking button". The "integration" part is already implied via the file name.

src/mocks/eventsforce/testDataBookingButton/activeSessionWithPrice.js
```diff
@@ -0,0 +1,308 @@
...
+      displayOrder: 3,
+    },
+    {
+      ticketType: 'group',
+      identifier: '40',
+      name: 'Family Ticket',
+      displayOrder: 2,
+    },
+  ],
+}
```
mpcarolin, line 308: Most of this is just duplicating the existing `testData.json` file's content. It would be better to do:
```js
import testData from 'SVEvfMocks/eventsforce/testData.js'

export default {
  ...testData,
 sessions: [
    {
      allowBooking: true,
      identifier: '0',
      name: 'Session Title',
      summary: 'Test summary',
      dayNumber: 1,
      startDateTimeLocal: '2020-08-03 09:00:00',
      endDateTimeLocal: '2020-08-03 13:30:00',
      presenterIdentifiers: [ '1', '2' ],
      labelIdentifiers: [ '1', '2', '3', '4' ],
      locationIdentifier: '1',
      liveVideoUrl: 'https://us02web.zoom.us/j/1234',
      recordedVideoUrl: 'https://www.youtube.com/watch?v=21X5lGlDOfg',
      restrictToAttendeeCategories: [],
      capacity: {
        isUnlimited: true,
      },
      price: {
        currency: 'USD',
        amount: 923.0,
      },
    },
  ],

```
(also note that I just made the "summary" field shorter, to be less noisy)

This is a lot cleaner I think, and if we ever need to change the `testData.json` contents, these tests will receive those updates.

mpcarolin: also, you can just name the folder "bookingButton", because all of the data in `mocks/eventsforce` is testData, so it's implied.

src/components/__tests__/bookingButton.integration.js
```diff
@@ -0,0 +1,41 @@
...
+  it('For an active session with price and NO waiting list : Should have state displaying BUY with price present', async () => {
```
mpcarolin, line 19: For each of these tests, it would be clearer to follow the jest conventions and use the `it('should <assertion>)` format for test names.

For example:
```js
  it('should display BUY <PRICE>, for active sessions with price and no waiting list', async () => {
```

mpcarolin (changes requested):
Hey it's looking good. These tests are SUPER helpful. They tell me how the booking button is supposed to behave, which I never had memorized.

I just have a few minor recommendations, mostly with naming.

src/components/__tests__/bookingButton.integration.js
subaya710: Done

src/mocks/eventsforce/testDataBookingButton/activeSessionWithPrice.js
subaya710: Done , Great suggestion!

src/components/__tests__/bookingButton.integration.js
subaya710: Done

mpcarolin (approved):
Looks good. Thank you!

PR #156, opened 2021-10-28
Add new action to reject outdated PRs

mpcarolin:
## Goal
* The dev team has agreed that no PR should be reviewed or QA-tested unless it is up to date with the `develop`
* This github action should automatically check that
* We only need to check for a green check-mark when a PR is opened

## Testing
* Verify the action ran on this PR and succeeded:

image

## Notes
* We should gradually add this action to any repo we work on


subaya710 (approved):
Approved. Looks good

PR #158, opened 2022-02-11
Group Modal Fixes and Tests

mpcarolin:
> Note: I will be branching off of this for ZEN-669, so please do not postpone review
## Context

* As I was working on `ZEN-669`, I noticed that we had actually broken the group booking modal in recent commits on develop.
* Furthermore, the group booker lacked any integration tests, so I wanted to correct that

## Updates

* `src/components/__tests__/groupBooker.integration.js`
  * Add some integration tests for the group booker. This is by no means exhaustive, but it's a good start.
  * Fix the bug in which the event handler for checkbox-click was not being passed the attendee id but rather the DOM element's id, which had additional text in it.
  * Wrapped the component in a `View` to that I could add some accessibility metadata
* `src/components/booking/attendeeCheckboxLabel.js`, `src/components/button/evfButton.js`
  * added an accessibility role
  * I used the `role` attribute, rather than `accessibilityRole`, because this component actually isn't a react-native one. The build exports a purely-web component (when running on web), so it's not react-native-web at all :| 
* `src/components/checkbox/evfCheckbox.js`
  * more bug fixing and better docs and better accessibility
* `types/efSessionAgenda.ts`
  * added the latest interface file

## Testing
* Ensure all the jest tests pass

### DIFF
--- a/src/components/__tests__/bookingButton.integration.js
+++ b/src/components/__tests__/bookingButton.integration.js
@@ -15,7 +15,7 @@ const getSelectBtnWithOnlyText = () =>
 const getWaitingListBtnWithOnlyText = () =>
   screen.getByRole('button', { name: 'WAITING LIST' })
 
-describe('Booking Button', () => {
+describe('Booking Button - Integration', () => {
   it('should display BUY <price> FOR an active session with price and NO waiting list.', async () => {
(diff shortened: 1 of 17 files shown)


subaya710 (approved):
Looks good . Approved

